PR #10, opened 2018-05-10
Updates for docsite inclusion

localredhead:
Site Generator updates to pull in this branch:  https://github.com/uport-project/uport-project.github.io/pull/20/files

The current structure looks like this when deployed with these changes:  
screen shot 2018-05-10 at 3 04 43 pm

I am still working on the content & information hierarchy, so things till change slightly from this screen shot, but I hope this helps add context to some of the changes.  Generally the changes were to try and include the information needed in the docs-site in a way that could be somewhat standardized.

The docs-repo contains some general guidelines for my approach here.  The following suggestions will help answer questions as to why I moved some content around, the rest of the guidelines are here:  https://github.com/uport-project/docs/blob/develop/README.md#repository-guidelines

```- Keep the readme short and sweet and detail how to contribute / support.
- Put detailed content, like implementation details in the /docs folder.
- Docs folders can have: guides, tutorials, sdk/api reference documentation.
```

I will change the suggestions above to also include basic install/getting started quick copy in the readme.

The generated documention needs a little work too, which requires some edits to the pre-generated code.  

Generated documentation (showing under reference):
screen shot 2018-05-11 at 1 07 19 am

This is the issue with the generated doc though... note the SimpleSigner => bit above....   
hopefully we can change this so the reference to simpleSigner is not so redundant... the links are a little weird.
screen shot 2018-05-11 at 1 10 16 am
screen shot 2018-05-11 at 1 10 10 am

The link in the nav ends up looking like:  `http://localhost:8000/did-jwt/reference/index#simple-signer-hex-private-key-%E2%87%92`


localredhead:
Update...  This is how things sit now.  I will need some input on the nav headings.
screen shot 2018-05-11 at 1 29 17 am
screen shot 2018-05-11 at 1 29 07 am

docs/reference/index.md
```diff
@@ -0,0 +1,159 @@
...
+| --- | --- | --- |
+| jwt | <code>String</code> | a JSON Web Token to verify |
+| [config] | <code>Object</code> | an unsigned credential object |
+| config.auth | <code>Boolean</code> | Require signer to be listed in the authentication section of the DID document (for Authentication purposes) |
+| config.audience | <code>String</code> | DID of the recipient of the JWT |
+| config.callbackUrl | <code>String</code> | callback url in JWT |
+
+**Example**  
+```js
+verifyJWT('did:uport:eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NksifQ.eyJyZXF1Z....', {audience: '5A8bRWU3F7j3REx3vkJ...', callbackUrl: 'https://...}).then(obj => {
```
oed: There is a missing `'` on this line

README.md
```diff
@@ -41,142 +23,74 @@ or if you use `yarn`
...
-Use the `createJWT()` function
+## Example
 
+### 1. Create a did-JWT
 ```js
-import { createJWT, SimpleSigner } from 'did-jwt'
-
-const signer = SimpleSigner('PRIVATEKEY')
+const didJWT = require('did-jwt')
+const signer = didJWT.SimpleSigner('fa09a3ff0d486be2eb69545c393e2cf47cb53feb44a3550199346bdfa6f53245');
```
oed: Maybe add a comment that this is just an example of a private key?

README.md
```diff
@@ -41,142 +23,74 @@ or if you use `yarn`
...
 ```
 
-## API
-
-### Creating a JWT
-
-Use the `createJWT()` function
+## Example
 
+### 1. Create a did-JWT
```
oed: Woud be good to have a link to the api reference for the function described.
Same thing for the `2. Verify a did-JWT` section.

README.md
localredhead: It's an actual identity private key so readers can cut/paste the example and run it.

How can we do this and not confuse the reader?

docs/reference/index.md
localredhead: Fixed!  Thanks for spotting it

README.md
localredhead: I added some text that clarifies the signer key bit.  LMK what you think, and if this is a good idea or should we do something different?

README.md
localredhead: Great suggestion!   We need to do this in many places, starting here :)

Fixed.

PR #15, opened 2018-08-01
Align documentation with code function parameters

### LINKED ISSUE #7: Align documentation name with argument name in verifyJWT
Minor thing, arg is called options, but documented as config, align them or could rewrite as `verifyJWT(jwt, {auth, audience, callbackUrl} = {})`

https://github.com/uport-project/did-jwt/blob/10df837635ef443bd9b54a982cfa5f8d2a844e58/src/JWT.js#L108

localredhead:
Closes #7

### DIFF
--- a/src/JWT.js
+++ b/src/JWT.js
@@ -67,10 +67,10 @@ export function decodeJWT (jwt) {
 *  })
 *
 *  @param    {Object}            payload            payload object
-*  @param    {Object}            [config]           an unsigned credential object
-*  @param    {String}            config.issuer      The DID of the issuer (signer) of JWT
-*  @param    {String}            config.alg         The JWT signing algorithm to use. Supports: [ES256K, ES256K-R], Defaults to: ES256K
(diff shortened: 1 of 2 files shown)


src/JWT.js
```diff
@@ -67,10 +67,10 @@ export function decodeJWT (jwt) {
 *  })
 *
 *  @param    {Object}            payload            payload object
-*  @param    {Object}            [config]           an unsigned credential object
-*  @param    {String}            config.issuer      The DID of the issuer (signer) of JWT
-*  @param    {String}            config.alg         The JWT signing algorithm to use. Supports: [ES256K, ES256K-R], Defaults to: ES256K
-*  @param    {SimpleSigner}      config.signer      a signer, reference our SimpleSigner.js
+*  @param    {Object}            [options]           an unsigned credential object
```
zachferland: This was just moved over, but I find the description confusing, also generally what args are require here and which are not? if required maybe we don't have them in an options object. We can consider in another PR instead of this one, if you want to merge in, changes look good.

PR #19, opened 2018-09-12
feat: export normalizeDID function from JWT.js

rmw2:
This is recommended solution for coercing legacy MNID addresses in JWTs to DIDs.  
We can export the function to make migration easier for integrators

To avoid polluting the exports from this library, we can then pass it through as an export of the minor release of `uport-credentials`
```javascript
// in credentials/index.js
import { normalizeDID } from 'did-jwt/lib/JWT'
// ...
export {Credentials, SimpleSigner, Contract, ContractFactory, JWT, normalizeDID } 
```

In particular this will help the TI&M team deal with legacy credentials

### DIFF
--- a/src/JWT.js
+++ b/src/JWT.js
@@ -27,7 +27,7 @@ function isDIDOrMNID (mnidOrDid) {
   return mnidOrDid && (mnidOrDid.match(/^did:/) || isMNID(mnidOrDid))
 }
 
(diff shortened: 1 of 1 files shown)


Eseoghene:
@pelle Is this okay to be merged

pelle:
It doesn't solve the problem that TI&M has, but it's probably useful.

PR #21, opened 2018-11-05
Update base64url to avoid vulnerabilities

ericmaino:
Base64url has buffer overrun vulnerabilities prior to 3.0.0. This change updates did-jwt to use 3.0.0 to avoid these vulnerabilities.


pelle:
Updated in other PR

PR #26, opened 2018-12-27
Update to latest resolver

pelle:
Also fixes various securities issues in dependencies


package.json
```diff
@@ -47,15 +48,15 @@
   },
   "dependencies": {
     "babel-runtime": "^6.26.0",
-    "base64url": "^2.0.0",
+    "base64url": "^3.0.1",
     "buffer": "^5.1.0",
-    "did-resolver": "^0.0.4",
+    "did-resolver": "0.0.6-alpha5",
```
oed: Looks like you just released 0.0.5. Why are you updating it to this instead?

pelle: New version solves a few problems that would occasionally happen with multiple instances of the REGISTRY existing with different versions of the library being loaded by npm.

It is also rewritten in Typescript, which makes its easier to write new did resolvers.

PR #34, opened 2019-04-12
remove non-test usages of nacl-did

mi-xu:


src/__tests__/JWT-test.js
```diff
@@ -166,45 +158,6 @@ describe('verifyJWT()', () => {
     })
   })
 
-  describe('nacl-did jwt', () => {
```
pelle: I don't think we should get rid of these tests? nacl-did is fine as a dev dependency. These tests test that it works with ED25519 signatures.

mi-xu: Done

PR #36, opened 2019-05-07
Convert to Typescript

mi-xu:


pelle:
@mi-xu there are failures

PR #37, opened 2019-05-15
feat: Enable JWT creation without timestamps

oed:
Need a way to create JWTs that don't contain a timestamp. This PR simply adds an option to do just that.


mirceanis:
This should already be possible by providing a `payload` object with the `"iat"` field set to `undefined` and the `expiresIn` parameter set to something falsey.

When the result payload is computed, there's a `{ ...timestamps, ...payload }` expression.
In the case of collision, the fields in `payload` should take precedence.

oed:
Oh, didn't think of that. Thanks @mirceanis. Closing this.

PR #38, opened 2019-05-16
feat: Expose Verifier- and SignerAlgorithm

oed:
Need the Verifier- and SignerAlgorithm separately. This PR exposes them in the exports.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,7 @@
 import SimpleSigner from './SimpleSigner'
 import NaclSigner from './NaclSigner'
 import { verifyJWT, createJWT, decodeJWT } from './JWT'
+import VerifierAlgorithm from './VerifierAlgorithm'
+import SignerAlgorithm from './SignerAlgorithm'
 
-module.exports = { SimpleSigner, NaclSigner, verifyJWT, createJWT, decodeJWT }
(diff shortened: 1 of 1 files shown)


mirceanis:
What is the purpose of this?
Those are internal implementations.
I'm not sure this should become public API, since it can't receive any support.

oed:
I basically want the ability to sign messages without encoding the data as a JWT format. It seems unlikely that these implementations will change, so I don't see any harm in exposing them. 

Most basically I want to be able to verify a signature without resolving a DID. I guess an option to pass an `authenticators` object to verifyJWT could also work.

mirceanis:
Even with the verifier alg exposed, you still need to construct a list of `authenticators` to pass around.
`verifyJWT` does not take that as a parameter.
What is the API you are looking to call?

It is not only about potentially changing the internal methods. Having this exposed here means we would have to provide similar methods on other platforms. The implementations there are completely different.

mi-xu:
@oed If you want to sign messages and verify those signatures without doing anything involving JWTs then does it make sense to do that through this library?  What about just using secp256k1 from the Elliptic module instead?

oed:
Managed to solve it by just taking the signature part of the JWT, then reconstructing the header + payload when verifying. 
My use case is basically that I already have the data that I'm signing in the data structure and just want to add a proof. So it doesn't make sense to also store the JWT payload.

PR #40, opened 2019-06-11
[WIP] Feature/166534105/fix global buffer

mi-xu:


package.json
```diff
@@ -80,14 +80,14 @@
...
-    "base64url": "^3.0.1",
+    "@stablelib/utf8": "^0.10.1",
     "buffer": "^5.2.1",
     "did-resolver": "0.0.6",
     "elliptic": "^6.4.0",
     "js-sha256": "^0.9.0",
     "js-sha3": "^0.8.0",
     "tweetnacl": "^1.0.1",
-    "tweetnacl-util": "^0.15.0"
+    "uport-base64url": "3.0.2-alpha.0"
```
oed: What's up with this? Anything broken upstream?

mi-xu: Dependencies that assume the presence of Buffer without explicitly importing it cause problems in react native.  This is to replace their implementations with ones that do.  Nothing should be broken upstream.

oed: Ok, is the intention to make a PR upstream? Seems troublesome if you have to maintain this fork for yourself?

localredhead: Yes we plan to open a PR upstream once we've done our rounds of integration testing

oed: Awesome ✌️

PR #41, opened 2019-07-25
change iat to nbf

mi-xu:
`did-jwt` currently sets the `iat` attribute to the current timestamp in seconds.  The W3C VC JWT spec has been changed to use the `nbf` attribute instead.  This PR changes all occurrences of `iat` to `nbf` when creating and verifying JWTs.

Tests have been updated to check for the presence of `nbf`.

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -43,7 +43,7 @@ interface JWTPayload {
   iss?: string
   sub?: string
   aud?: string
-  iat?: number
+  nbf?: number
   type?: string
   exp?: number
   rexp?: number
@@ -88,7 +88,7 @@ function encodeSection(data: any): string {
   return base64url.encode(JSON.stringify(data))
 }
 
-export const IAT_SKEW: number = 300
+export const NBF_SKEW: number = 300
 
 /**  @module did-jwt/JWT */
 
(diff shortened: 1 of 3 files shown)


awoie:
to be more backward compatible, you could use both `iat` and `nbf`. `iat` is not in the spec but allowed.

mirceanis:
Agree with @awoie but since there is no guidance (that I know of) on how to interpret the fields during validation, other than semantics, it would be OK like this too.

This is a breaking change and must be reflected in the version when released

oed:
We are currently setting `iat` to `null` in order to make the output of a JWT deterministic in some cases. This change could possibly break our production code. Please interpret `iat` as `nbf` if passed in the payload.

awoie:
There is some guidance for `nbf`. Note, that `iat` by itself does not require validation. 

From the JWT spec (https://tools.ietf.org/html/rfc7519#section-4.1.5, https://tools.ietf.org/html/rfc7519#section-4.1.6):

**`iat`**
> The "iat" (issued at) claim identifies the time at which the JWT was
> issued.  This claim can be used to determine the age of the JWT.  Its
> value MUST be a number containing a NumericDate value.  Use of this
> claim is OPTIONAL.

**`nbf`**
> The "nbf" (not before) claim identifies the time before which the JWT
> MUST NOT be accepted for processing.  The processing of the "nbf"
> claim requires that the current date/time MUST be after or equal to
> the not-before date/time listed in the "nbf" claim.  Implementers MAY
> provide for some small leeway, usually no more than a few minutes, to
> account for clock skew.  Its value MUST be a number containing a
> NumericDate value.  Use of this claim is OPTIONAL.

mi-xu:
@oed You can still set `iat` or `nbf` to null by providing it in your `payload`.  It would be semantically incorrect to interpret `iat` from the `payload` as `nbf`.

PR #42, opened 2019-07-29
[WIP] potentially breaking changes needed for W3C VC support

mi-xu:
In this PR:
- allow creation of a signed JWT without specifying issuer did
- allow creation of unsigned did

Other potential changes:
- remove logic for checking `aud` from `verifyJWT`
- API of `createJWT`
- automatically populated attributes (`iat`, `nbf`, etc)?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "did-jwt",
-  "version": "0.2.0",
+  "version": "1.0.0-dev.0",
   "description": "Library for Signing and Verifying JWTs compatible uPort and DID standards",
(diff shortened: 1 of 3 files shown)


mi-xu:
move `createUnsignedJWT` to `uport-vc-driver`

awoie:
we should keep the `iss` as mandatory although it is not mandatory for W3C presentations. otherwise, we won't be able to verify the JWT. i don't see an issue with having the `iss` as part of the `payload` itself but i am not able to assess if there are other implications in other parts of the lib or dependent libs.

PR #43, opened 2019-08-12
Set `iat` timestamp by default instead of `nbf`

mi-xu:
This PR reverts the changes to `createJWT` introduced in https://github.com/uport-project/did-jwt/pull/41 and slightly alters the behavior of how some of the arguments are used:

- `createJWT` no longer attempts sets the `nbf` attribute to the current timestamp by default
- `createJWT` once again sets the `iat` attribute to the current timestamp by default
- `createJWT` omits the `iat` attribute if it is passed in as `undefined` in the `payload`
- `createJWT` only calculates the `exp` value given the `expiresIn` option if `nbf` is present in the `payload`

`nbf` no longer defaults to the current timestamp is because it's an optional attribute for JWTs, and therefore does not justify causing a breaking change in this library.  `did-jwt-vc` will provide higher level functions for creating Verifiable Credentials and Presentations which are responsible for defaulting the value of `nbf` since it is a necessary attribute for verification in those contexts.

Another slight difference in behavior here is that the `expiresIn` option will be ignored if the `payload` does not specify `nbf`.  The previous behavior would only function appropriately in the scenario that `nbf` was not passed in as part of the `payload` since it would always calculate `exp` from the automatically generated `nbf` even if a value was assigned explicitly.

### DIFF
--- a/src/__tests__/JWT-test.ts
+++ b/src/__tests__/JWT-test.ts
@@ -186,7 +186,7 @@ describe('createJWT()', () => {
 
     it('creates a JWT with expiry in 10000 seconds', () => {
       return createJWT(
-        { requested: ['name', 'phone'], nbf: Math.floor(new Date().getTime()/1000) },
+        { requested: ['name', 'phone'], nbf: Math.floor(new Date().getTime() / 1000) },
         { alg, issuer: did, signer, expiresIn: 10000 }
       ).then(jwt => {
         const { payload } = decodeJWT(jwt)
--- a/src/__tests__/JWT-test.ts
+++ b/src/__tests__/JWT-test.ts
@@ -107,7 +107,7 @@ describe('createJWT()', () => {
 
     it('creates a JWT with expiry in 10000 seconds', () => {
       return createJWT(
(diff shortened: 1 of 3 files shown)


src/__tests__/JWT-test.ts
```diff
@@ -107,14 +107,48 @@ describe('createJWT()', () => {
 
     it('creates a JWT with expiry in 10000 seconds', () => {
       return createJWT(
-        { requested: ['name', 'phone'] },
+        { requested: ['name', 'phone'], nbf: Math.floor(new Date().getTime()/1000) },
```
mirceanis:
```suggestion
        { requested: ['name', 'phone'], nbf: Math.floor(new Date().getTime() / 1000) },
```

src/__tests__/JWT-test.ts
```diff
@@ -152,7 +186,7 @@ describe('createJWT()', () => {
...
```
mirceanis:
```suggestion
        { requested: ['name', 'phone'], nbf: Math.floor(new Date().getTime() / 1000) },
```

PR #45, opened 2019-08-13
feat: verifyJWT against nbf if present but use iat if not

mi-xu:
This PR adds support for checking the current time against the `nbf` field when verifying a JWT.  It prioritizes the presence of `nbf` in order to comply with W3C VC standards, but falls back to `iat` when it is absent in order to maintain support for legacy format JWTs.

This completes [#167881436](https://www.pivotaltracker.com/n/projects/2198688/stories/167881436) on pivotal

Test scenarios have been added to validate the behavior for all combinations of `nbf` and `iat` fields being part of the JWT or not.

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -233,7 +233,7 @@ export async function verifyJWT(
     } else if (payload.iat && payload.iat > nowSkewed) {
       throw new Error(`JWT not valid yet (issued in the future) iat: ${payload.iat}`)
     }
-    if (payload.exp && payload.exp <= nowSkewed) {
+    if (payload.exp && payload.exp <= now - NBF_SKEW) {
       throw new Error(`JWT has expired: exp: ${payload.exp} < now: ${now}`)
     }
     if (payload.aud) {
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -225,14 +225,15 @@ export async function verifyJWT(
   )
   const now: number = Math.floor(Date.now() / 1000)
   if (signer) {
-    if (payload.nbf && payload.nbf > now + NBF_SKEW) {
-      throw new Error(
-        `JWT not valid yet (issued in the future): nbf: ${
-          payload.nbf
-        } > now: ${now}`
-      )
+    const nowSkewed = now + NBF_SKEW
+    if (payload.nbf) {
+      if (payload.nbf > nowSkewed) {
+        throw new Error(`JWT not valid before nbf: ${payload.nbf}`)
+      }
+    } else if (payload.iat && payload.iat > nowSkewed) {
+      throw new Error(`JWT not valid yet (issued in the future) iat: ${payload.iat}`)
     }
-    if (payload.exp && payload.exp <= now - NBF_SKEW) {
+    if (payload.exp && payload.exp <= nowSkewed) {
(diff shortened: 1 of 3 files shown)


src/__tests__/JWT-test.ts
```diff
@@ -297,14 +299,56 @@ describe('verifyJWT()', () => {
...
+    it('fails when nbf is missing and iat is in the future', async () => {
+      const jwt = await createJWT(
+        { iat: FUTURE },
+        { issuer: did, signer }
+      )
+      expect(verifyJWT(jwt)).rejects.toThrow()
+    })
+    it('passes when nbf and iat are both missing', async () => {
+      const jwt = await createJWT(
+        {},
```
mirceanis: The jwt created here still has an `iat` field added by default.
We should probably use `"iat" : undefined` in the payload to force it to be removed, or even better, use hard-coded JWTs for these tests.

src/__tests__/JWT-test.ts
```diff
@@ -331,25 +375,12 @@ describe('verifyJWT()', () => {
...
-          )
-          .then(p => expect(p).toBeFalsy())
-    )
-  })
-
   it('accepts a valid exp', () => {
     return createJWT(
-      { exp: NOW - NBF_SKEW + 1 },
-      { issuer: did, signer, expiresIn: 1 }
+      { exp: NOW + 1000 },
```
mirceanis: The NBF_SKEW should be the leeway to account for clock skew. That means that a token that expires now should still be considered valid for another NBF_SKEW seconds.

I believe this test should cover something closer to this scenario. `exp: NOW` should work.

src/JWT.ts
```diff
@@ -225,14 +225,15 @@ export async function verifyJWT(
...
+    const nowSkewed = now + NBF_SKEW
+    if (payload.nbf) {
+      if (payload.nbf > nowSkewed) {
+        throw new Error(`JWT not valid before nbf: ${payload.nbf}`)
+      }
+    } else if (payload.iat && payload.iat > nowSkewed) {
+      throw new Error(`JWT not valid yet (issued in the future) iat: ${payload.iat}`)
     }
-    if (payload.exp && payload.exp <= now - NBF_SKEW) {
+    if (payload.exp && payload.exp <= nowSkewed) {
```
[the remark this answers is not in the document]
mi-xu: fixed

src/__tests__/JWT-test.ts
mi-xu: fixed

src/__tests__/JWT-test.ts
mi-xu: fixed

src/__tests__/JWT-test.ts
```diff
@@ -20,6 +20,8 @@ registerResolver()
 registerNaclDID()
 
 const NOW = 1485321133
+const PAST = NOW - 60000
```
mirceanis: codacy is complaining about this
```suggestion
```

src/__tests__/JWT-test.ts
```diff
@@ -20,6 +20,8 @@ registerResolver()
...
+const FUTURE = NOW + 60000
```
mirceanis: codacy is complaining about this too
```suggestion
```

PR #46, opened 2019-08-14
release: add support for nbf attribute

mi-xu:
This PR releases the changes to support using the `nbf` field to represent the beginning of the JWTs validity period.  JWTs created before this change that use `iat` will still be supported as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,13 +26,48 @@ jobs:
           name: code-coverage
           command: bash <(curl -s https://codecov.io/bash)
 
+      - run:
+          name: commitlint
+          command: |
+            echo "module.exports = {extends: ['@commitlint/config-conventional']}" > commitlint.config.js
+            npm install @commitlint/cli @commitlint/config-conventional
+            npx semantic-commitlint -- --ci
+
       - save_cache:
           key: dependency-cache-{{ checksum "package.json" }}
           paths:
             - ./node_modules
+
+  release:
+    working_directory: ~/did-jwt
+    docker:
+      - image: circleci/node:8
+    steps:
+      - add_ssh_keys:
+          fingerprints:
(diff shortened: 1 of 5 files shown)


mirceanis:
I think we should come up with a prefix for these release triggers.
Maybe use `[Release]` as the PR title prefix?

I got confused seeing the same changeset here as the previous PR

.circleci/config.yml
```diff
@@ -26,13 +26,48 @@ jobs:
           name: code-coverage
           command: bash <(curl -s https://codecov.io/bash)
 
+      - run:
+          name: commitlint
+          command: |
+            echo "module.exports = {extends: ['@commitlint/config-conventional']}" > commitlint.config.js
```
mirceanis: If we decide to use this tool, it's probably easier to check this file into git instead of echoing here

.circleci/config.yml
```diff
@@ -26,13 +26,48 @@ jobs:
...
+            npm install @commitlint/cli @commitlint/config-conventional
```
mirceanis: if we decide to use this tool, it's probably better to add these as devDependencies in package.json instead of installing them here every time

.circleci/config.yml
```diff
@@ -26,13 +26,48 @@ jobs:
...
+            npx semantic-commitlint -- --ci
```
mirceanis: What's the action that a developer can take if this command fails?
How exactly does this help?

.circleci/config.yml
localredhead: I want the config here to be agnostic of NPM projects hence why it is wired this way

.circleci/config.yml
localredhead: They rebase to update the commit.   https://help.github.com/en/articles/changing-a-commit-message

.circleci/config.yml
localredhead: It's done this way on purpose to remain agnostic of npm vs. non npm projects

PR #56, opened 2019-09-04
feat: add getter and setter for did resolver

mi-xu:
This PR adds two functions `getResolver()` and `setResolver()` that allow the user to override the did-resolver used by `verifyJWT()`.  By default, it will still use a did-resolver that has ethr-did and https-did resolvers configured without any customizations.

Although this implementation is different from what was previously discussed of turning did-jwt into a class and having the resolver be passed as an optional param of the constructor, I believe that using a setter instead can achieve the same effect with much fewer (and no breaking) changes.  Additionally, refactoring it to be encapsulated within a class needlessly changes the way all other functions are accessed when `verifyJWT()` is the only one that uses the resolver.

@mirceanis what do you think about this approach?

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -86,10 +86,18 @@ const SUPPORTED_PUBLIC_KEY_TYPES: PublicKeyTypes = {
 
 const defaultAlg = 'ES256K'
 
-export const resolver = new Resolver({
+// tslint:disable-next-line: variable-name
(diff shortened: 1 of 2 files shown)


mirceanis:
The class approach better isolates the resolver used.

If a library decides to expose its own JWT verification method with a custom resolver it will break an entire app since the setter works on a global.

Issue #65, opened 2019-10-29
`JWTPayload` is filtering out custom JWT properties


mirceanis:
**Describe the bug**

`decodeJWT` returns a `JWTPayload` which defines a strict set of properties it may contain.
When a JWT with custom properties is being decoded then the decoded payload will not see these properties.

**To Reproduce**

use `did-jwt` in a typescript project, like so:
```typescript
import { decodeJWT } from 'did-jwt'

//a JWT with a `status` property in the payload
const decoded = decodeJWT('eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NkstUiJ9.eyJpYXQiOjE1NzIzNTE2NTYsInN0YXR1cyI6eyJpZCI6InJpbmtlYnk6MHhTdGF0dXNSZWdpc3RyeUFkZHJlc3MiLCJ0eXBlIjoiRXRoclN0YXR1c1JlZ2lzdHJ5MjAxOSJ9LCJpc3MiOiJkaWQ6ZXRocjoweGYzYmVhYzMwYzQ5OGQ5ZTI2ODY1ZjM0ZmNhYTU3ZGJiOTM1YjBkNzQifQ.OoR164SxkZEl0Q0aW7yBkKpAnWL4DYeiBtN85L4tD793mfhxb_Q2rxPot0DUFXVnBULoS55x5MYuRMVRpOSl7QE')

console.log(decoded.payload.status)
// compilation fails with` Property 'status' does not exist on type 'JWTPayload'.`
// a workaround is to cast to `any`
```

**Expected behavior**

`JWTPayload` and `JWTHeader` should allow custom properties.
Compilation should not fail in the scenario above

**Environment**

- did-jwt version 3.0.0

PR #73, opened 2020-04-02
cty header in jwt

nikolockenvitz:
I need to add the `cty` header parameter in one of my applications, so I added the possibility to add it when creating a JWT. (See [RFC](https://tools.ietf.org/html/rfc7519#section-5.2))

One could also think of a parameter in the createJWT()-method to specify custom headers which get injected into the JWT header, but the cty header alone would be enough for my use case right now.

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -42,6 +42,7 @@ interface DIDAuthenticator {
 interface JWTHeader {
   typ: 'JWT'
   alg: string
+  cty?: 'JWT'
 }
 
 interface JWTPayload {
@@ -158,11 +159,13 @@ export function decodeJWT(jwt: string): JWTDecoded {
 // export async function createJWT(payload, { issuer, signer, alg, expiresIn }) {
 export async function createJWT(
   payload: any,
-  { issuer, signer, alg, expiresIn }: JWTOptions
+  { issuer, signer, alg, expiresIn }: JWTOptions,
+  cty?: boolean
 ): Promise<string> {
   if (!signer) throw new Error('No Signer functionality has been configured')
   if (!issuer) throw new Error('No issuing DID has been configured')
-  const header: JWTHeader = { typ: 'JWT', alg: alg || defaultAlg }
+  let header: JWTHeader = { typ: 'JWT', alg: alg || defaultAlg }
+  if (cty) header.cty = 'JWT'
   const timestamps: Partial<JWTPayload> = {
     iat: Math.floor(Date.now() / 1000),
     exp: undefined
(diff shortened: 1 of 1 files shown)


mirceanis:
This kind-of relates to #65 in the sense that the data types for `JWTHeader` and `JWTPayload` are not extensible.

nikolockenvitz:
Ok, so should I add that custom headers are working in general (when creating and decoding)?

mirceanis:
I'm picturing these changes to cover all of the above:

* An extension to `JWTHeader`
```typescript
interface JWTHeader {
  typ: 'JWT'
  alg: string
  [x: string]: any
}
```
* A similar extension to `JWTPayload`
```typescript
interface `JWTPayload` {
  iss?: string
  sub?: string
  aud?: string
  iat?: number
  nbf?: number
  type?: string
  exp?: number
  rexp?: number
  [x: string]: any
}
```
* A change to the `createJWT` signature, to add an optional `header` param
```typescript
export async function createJWT(
  payload: any,
  { issuer, signer, alg, expiresIn }: JWTOptions,
  header: JWTHeader,
): Promise<string> {	)
```
The `alg` param in `JWTOptions` would be deprecated in favor of a fully specified `JWTHeader`

nikolockenvitz:
I updated the function so that all existing implementations should work with it too:
```ts
export async function createJWT(
  payload: any,
  { issuer, signer, alg, expiresIn }: JWTOptions,
  header: Partial<JWTHeader> = {},
): Promise<string> {
  if (!signer) throw new Error('No Signer functionality has been configured')
  if (!issuer) throw new Error('No issuing DID has been configured')
  if (!header.typ) header.typ = 'JWT'
  if (!header.alg) header.alg = alg || defaultAlg
```

mirceanis:
It looks great.
There is one more step to getting this merged; fixing the commit messages.

This repository is using [semantic-release](https://github.com/semantic-release/semantic-release) and release and versioning decisions are based on commit messages; that is why the [`ci/circleci: build`](https://circleci.com/gh/decentralized-identity/did-jwt/184) step is failing.

In your case, a commit message that would fit this PR would be something like this:
```
feat: add parameter for custom headers to createJWT

also fixes #65
```

nikolockenvitz:
So I just make an additional commit with such a message, right? I can check whether I need to update the docs

mirceanis:
> So I just make an additional commit with such a message, right? I can check whether I need to update the docs

 I think `commitlint` would fail if you only added a new commit but kept the previous ones with that format.
Since the first commit is effectively overwritten by the second, I think it's best to recreate this PR as a single commit with the proper message.

---

Alternatively you could split it into 2 commits:

```
fix: extend data types for JWTHeader and JWTPayload

fixes #65
```

and

```
feat: add parameter for custom headers to `createJWT`
```

nikolockenvitz:
Thanks, I just recreate the PR with one commit

nikolockenvitz:
I create a new PR #76

PR #84, opened 2020-04-24
feat: add methods for creating and verifying JWS

oed:
This PR add support to create and verify plain JWS. It adds two new functions `createJWS` which is now used by the `createJWT` function, and `verifyJWS` which is **not** used by `verifyJWT` (not much overlap).

In addition I also made some small enhancements:
* Updated the test script to not automatically update the snapshots
* Made the signature verification error messages more informative by adding the token that was failed to verify

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "esm"
   ],
   "scripts": {
-    "test": "./node_modules/.bin/standard && jest --updateSnapshot",
+    "test": "./node_modules/.bin/standard && jest",
     "build:js": "./node_modules/.bin/microbundle",
     "build:browser": "./node_modules/.bin/webpack --config webpack.config.js",
     "build": "npm run build:js && npm test && npm run build:browser",
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -142,6 +142,30 @@ export function decodeJWT(jwt: string): JWTDecoded {
 }
 
 /**
+ *  Creates a signed JWS given a payload, a signer, and an optional header.
+ *
+ *  @example
+ *  const signer = SimpleSigner(process.env.PRIVATE_KEY)
+ *  const jws = await createJWS({ my: 'payload' }, signer)
+ *
+ *  @param    {Object}            payload           payload object
+ *  @param    {SimpleSigner}      signer            a signer, reference our SimpleSigner.js
+ *  @param    {Object}            header            optional object to specify or customize the JWS header
+ *  @return   {Promise<Object, Error>}              a promise which resolves with a JWS string or rejects with an error
+ */
+export async function createJWS (payload: any, signer: Signer, header: Partial<JWTHeader> = {}): Promise<string> {
+  if (!header.alg) header.alg = defaultAlg
+  const signingInput: string = [
+    encodeSection(header),
+    encodeSection(payload)
+  ].join('.')
+
+  const jwtSigner: SignerAlgorithm = SignerAlgorithm(header.alg)
+  const signature: string = await jwtSigner(signingInput, signer)
+  return [signingInput, signature].join('.')
+}
+
+/**
  *  Creates a signed JWT given an address which becomes the issuer, a signer, and a payload for which the signature is over.
  *
  *  @example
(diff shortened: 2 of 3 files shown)


src/VerifierAlgorithm.ts
```diff
@@ -50,7 +54,7 @@ export function verifyES256K(
     signer = verifyRecoverableES256K(data, signature, ethAddressKeys)
   }
 
-  if (!signer) throw new Error('Signature invalid for JWT')
+  if (!signer) throw new Error(formatErrorMessage(data, signature))
```
mirceanis, line 57: Error messages should not indicate the data being verified.
These errors can bubble up and leak information.

src/JWT.ts
```diff
@@ -180,14 +203,29 @@ export async function createJWT(
...
+ *  Verifies given JWS. If the JWS is valid, returns the public key that was used to sign the JWS.
+ *
+ *  @example
+ *  const pubkey = verifyJWT('eyJ0eXAiOiJKV1QiLCJhbGciOiJFUzI1NksifQ.eyJyZXF1Z....', { publicKeyHex: '0x12341...' })
+ *
+ *  @param    {String}                          jws         A JWS string to verify
+ *  @param    {Array<PublicKey> | PublicKey}    pubkeys     The public keys used to verify the JWS
+ *  @return   {PublicKey}                       The public key used to sign the JWS
+ */
+export function verifyJWS (jws: string, pubkeys: PublicKey | PublicKey[]) {
```
mirceanis, line 220: The return type is missing, please mention `PublicKey`
```suggestion
export function verifyJWS (jws: string, pubkeys: PublicKey | PublicKey[]): PublicKey {
```

Also, this should be used in the `verifyJWT` method to not duplicate functionality.
I think it fits instead of the direct call to `VerifierAlgorithm` at L271
```
  const signer: PublicKey = VerifierAlgorithm(header.alg)(
    data,	    data,
    signature,	    signature,
    authenticators	    authenticators
  )	  )
```

src/JWT.ts
```diff
@@ -180,14 +203,29 @@ export async function createJWT(
...
-  ].join('.')
+  const fullPayload = { ...timestamps, ...payload, iss: issuer }
+  return createJWS(fullPayload, signer, header)
+}
 
-  const jwtSigner: SignerAlgorithm = SignerAlgorithm(header.alg)
-  const signature: string = await jwtSigner(signingInput, signer)
-  return [signingInput, signature].join('.')
+/**
+ *  Verifies given JWS. If the JWS is valid, returns the public key that was used to sign the JWS.
```
mirceanis, line 211:
```suggestion
 *  Verifies given JWS. If the JWS is valid, returns the public key that was used to sign the JWS. or throws an `Error` if none of the `pubkeys` match
```

package.json
```diff
@@ -15,7 +15,7 @@
     "esm"
   ],
   "scripts": {
-    "test": "./node_modules/.bin/standard && jest --updateSnapshot",
+    "test": "./node_modules/.bin/standard && jest",
```
mirceanis, line 18: nice catch!

src/VerifierAlgorithm.ts
oed: Reason I did this was because as I've been using the library the error messages are quite unhelpful. Thought this would help with that :)

src/JWT.ts
oed: Yeah, I looked at doing this, but it would result in two calls to `decodeJWT`. I can change it if thats ok though.

mirceanis: 2 calls to decode is not a big problem in the long run; especially since decoding is fairly simple and does not involve any external calls.
The risk with code duplication is that if one side changes it becomes inconsistent.

src/VerifierAlgorithm.ts
mirceanis: right, but that can also be done at an application level.
In your situation, wherever you call the method you know your JWT so it would be easier to catch the error and display it in a safe manner to aid your debugging.

The same cannot be said about all applications using this lib and since it's a common mistake to bubble up errors it can easily lead to information leaks.

oed:
Pushed changes based on your comments @mirceanis!

mirceanis:
looks great, thank you!

Issue #78, opened 2020-04-09
Clarification on signer methods and return types


mirceanis:
`signer` params are very confusing at the moment.

ES256K signers are expected to produce ECDSASignature objects while Ed25519 signers are expected to produce strings.

The developer is notified of this by an error.
See [SignerAlgorithm.ts#L28](https://github.com/decentralized-identity/did-jwt/blob/19572b671341053ab15ea311efdba8cdc4c867b6/src/SignerAlgorithm.ts#L28) and [SignerAlgorithm.ts#L41](https://github.com/decentralized-identity/did-jwt/blob/19572b671341053ab15ea311efdba8cdc4c867b6/src/SignerAlgorithm.ts#L41)

This should be improved.
Signers should only produce strings since that way, signing can be externalized and there would no longer be a need to understand all algorithms locally, thus allowing support for many more crypto suites.

oed:
Any thoughts on using a more generic underlying library such as [jose](https://github.com/panva/jose)? Would extend the number of supported algorithms :) 
Don't know how difficult it would be to migrate, but would be nice with wider support!

mirceanis:
Yes, this relates directly to that approach.

If this lib makes assumptions about the structure of signatures then it essentially prohibits the use of libs like `jose`

oed:
Is that because of the use of `ES256K-R`? Basically because of ethereum addresses being valid signature keys in this library?

mirceanis:
Yes, so there would still be special handling of non-standard algorithms; at least for a while.
The point of this issue is to relax the need of this library to know signature formats and expect signers to always produce strings.

Regarding ES256K-R, in the long run it should not even be a thing since, even for ethereumAddress as keys, it can be verified almost the same as ES256K.

In parallel, I made [a proposal to ethr-did-resolver](https://github.com/decentralized-identity/ethr-did-resolver/issues/56) to use public keys as identifiers instead of ethereum addresses. That would alleviate the need to have special handling here.

oed:
That sounds great :) FYI, we are likely to start using the `jose` library to do IPLD object signatures and encryption in Ceramic. But afaik, it should still be possible to verify signatures generated from that with the did-jwt library!

PR #90, opened 2020-04-28
fix: make SimpleSigner return a string

oed:
This PR addresses #78. Now both the NaclSigner and the SimpleSigner (secp256k1) returns JWS signatures. 
The Signer interface and SignerAlgorithm is still backwards compatible but will warn the user that they are using a deprecated version if they return the ecdsa object.

SimpleSigner now returns a recoverable signature by default. The recovery param will be removed if `ES256K` alg is being used.


oed:
Codacy is complaining about the `console.warm` happy for input on what to use instead!

PR #95, opened 2020-06-08
Refactor JWT.ts

ilanolkies:
Avoid decoding jws twice in verification


ilanolkies:
> SemanticReleaseError: The commit message is not formatted correctly

How should I do it?

PR #114, opened 2020-06-18
feat: add EllipticSigner that returns string

oed:
This PR adds a new signer `EllipticSinger` which is similar to `SimpleSigner` but returns a string instead of a signature object. It also adds support for string signatures in `SignerAlgorithms` when using `ES256K`.

Replaces #90


oed:
Thanks @mirceanis 🙏

PR #121, opened 2020-08-14
export more JWT interfaces

PeterTheOne:
Some of them may be useful in other libs like `uport-credentials` or `did-jwt-vc`.


mirceanis:
Thanks for the addition.
We haven't exported them until now because there was no direct need and because after it's public API, it needs to be maintained.

Since the main export is not the JWT.ts file I guess it would be ok to accept this PR

PR #126, opened 2020-08-18
feat: enable detached payloads for JWS

oed:
This PR enables a caller of `createJWS` to pass an `base64url` encoded payload directly. This is useful when you need to sign something that isn't a JSON object.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
   },
   "author": "Pelle Braendgaard <pelle.braendgaard@consensys.net>",
   "contributors": [
-    "Mircea Nistor <mircea.nistor@consensys.net>"
+    "Mircea Nistor <mircea.nistor@consensys.net>",
+    "Joel Thorstensson <oed@3box.io>"
   ],
   "license": "Apache-2.0",
   "jest": {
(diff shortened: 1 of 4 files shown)


mirceanis (changes requested):
It looks great but it needs some new tests.

* `decodeJWS` works as expected with a JWS input and fails as expected with garbage input
* `createJWS` works with both JSON payload and `string` payload (not sure if the method should fail when given an array as payload)

oed:
Thanks @mirceanis I'll add the tests!

> (not sure if the method should fail when given an array as payload)

I don't see why it should as that would be valid JSON.

oed:
> and fails as expected with garbage input

What do you mean by garbage input btw?

mirceanis:
> I don't see why it should as that would be valid JSON.

I can't find a concrete reference about the JSON payload of JWT but since the RFC talks about claims in the payload that can only happen if the payload is an object, not an array.

With this changeset, I suppose that typescript would complain if you tried to call the method with an array, or number instead of an object(because of `Record<string, any>`) which doesn't happen when you call it from javascript.
Having an explicit check for non-object types would make this more consistent, but I'm not sure if the JWT RFC agrees with this type of check.

oed:
In that case it seems like the check should be in the `createJWT` method and not the `createJWS` methods as JWS allows for an arbitrary payload.

oed:
That also seems out of scope for this PR?

oed:
Pushed an update with tests.

mirceanis:
> That also seems out of scope for this PR?

Yes, I was only pointing it out because of this new corner case inconsistency.

mirceanis (approved):
Looks great!

oed:
Thanks :pray:

PR #129, opened 2020-08-31
chore: reduce package size by not using tweetnacl

oed:
This PR replaces `tweetnacl` with `@stablelib/ed25519` which will reduce the package size. Still keeping `tweetnacl` as a devDep as it's used in a few tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,19 +86,20 @@
     "tslint": "6.1.3",
     "tslint-config-prettier": "1.18.0",
     "tslint-eslint-rules": "5.4.0",
+    "tweetnacl": "^1.0.3",
     "typescript": "4.0.2",
     "webpack": "4.44.1",
     "webpack-cli": "3.3.12"
   },
   "dependencies": {
     "@babel/runtime": "^7.11.2",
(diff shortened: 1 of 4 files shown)


mirceanis (changes requested):
This is great!

There is a failing build step because of the outdated `yarn.lock` file.
Please also update the lockfile.

Also, do you have a comparison between package sizes? How much is it reduced?

oed:
Ah, I don't use yarn. That might be why!

The difference is actually much less than I anticipated:
https://bundlephobia.com/result?p=tweetnacl@1.0.3
https://bundlephobia.com/result?p=@stablelib/ed25519@1.0.1

mirceanis:
> Ah, I don't use yarn. That might be why!
 
Normally, we wouldn't use `yarn` here either but there were some issues for which a workaround required `yarn`(see https://github.com/developit/microbundle/issues/657 and #101 )

I haven't checked if the issues were fixed in the meantime to switch back to vanilla npm

mirceanis (approved):

PR #131, opened 2020-09-11
fix: use EdDSA as the 'alg' header for Ed25519 signatures

### LINKED ISSUE #130: Incorrect "alg" header for Ed25519 signatures
**Describe the bug**
RFC 8037 requires that the "alg" header of a JOSE object be "EdDSA" when using Ed25519 signatures, did-jwt uses "Ed25519"

**To Reproduce**

The following code will throw an "Unsupported algorithm EdDSA" error:

```
import  didJWT from "did-jwt"

const signer = didJwt.NaclSigner(<some base64 bytes>)
const jws = await didJWT.createJWS({ asfd: 234 }, signer, {"alg": "EdDSA"})
```

**Expected behavior**
The JWS should be a valid JWS with the "EdDSA" alg header.

alexjg:
Fixes #130 

I've left the old algorithm parameters in for backwards compatibility. I don't know if this is desirable?

### DIFF
--- a/src/SignerAlgorithm.ts
+++ b/src/SignerAlgorithm.ts
@@ -36,7 +36,10 @@ interface SignerAlgorithms {
 const algorithms: SignerAlgorithms = {
   ES256K: ES256KSigner(),
   'ES256K-R': ES256KSigner(true),
-  Ed25519: Ed25519Signer()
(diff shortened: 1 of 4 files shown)


mirceanis (approved):
Looks good. Thank you!

PR #132, opened 2020-09-13
Add support for low level JWE functions

oed:
This PR adds support for low level JWE creation and decryption using the `XChacha20Poly1305` symmetric cipher and asymmetric cipher using `x25519` key exchange. It also includes a minimal implementation of the `concatKDF`. All crypto is implemented using the `@stablelib` packages.

Test vectors where generated using the [jose-chacha](https://github.com/panva/jose-chacha) library.

In addition all reliance on `Buffer` is removed from the library.


oed:
Merged master into this branch, so should be mergeable now.

mirceanis (approved):
It looks great. Thanks for this contrib.

I can imagine a few lines of work that are opened up now that the basis for JWE is here.
I'll ask for your review when needed :)

oed:
Thank you very much @mirceanis! Looking forward to future updates :)

Issue #127, opened 2020-08-19
Support base58 encoded public keys


oed:
Currently this library supports `publicKeyHex` for `ES256K` and `publicKeyBase64` for `x25519`. However the [did spec](https://w3c.github.io/did-core/#key-types-and-formats) seems to be converging on the use of `publicKeyBase58` everywhere. 

**Describe the solution you'd like**
The library should not assume that a specific key has a specific encoding. The most reasonable thing here seems to be to add support for all of the three encodings above for all key types.

Issue #128, opened 2020-08-20
[enhancement] use unified representation of key material


mirceanis:
As a prerequisite to fixing #127, this library needs to first transform public key material from DID docs to a unified representation that can later be converted to the needs of each verification method.

some code that will be affected:
https://github.com/decentralized-identity/did-jwt/blob/5ff35069ef7849f7e9721b07bafce9ce4cee18be/src/VerifierAlgorithm.ts#L30
https://github.com/decentralized-identity/did-jwt/blob/5ff35069ef7849f7e9721b07bafce9ce4cee18be/src/VerifierAlgorithm.ts#L37
https://github.com/decentralized-identity/did-jwt/blob/5ff35069ef7849f7e9721b07bafce9ce4cee18be/src/VerifierAlgorithm.ts#L73
https://github.com/decentralized-identity/did-jwt/blob/5ff35069ef7849f7e9721b07bafce9ce4cee18be/src/VerifierAlgorithm.ts#L91

PR #139, opened 2020-11-08
Fix/support pubkey encodings

oed:
This PR adds support for multiple public key encodings in the DID document. 
Fixes https://github.com/decentralized-identity/did-jwt/issues/127

Also removes the `@stablelib/utf8` dependency.


mirceanis (approved):
This looks great, @oed !
I think it also fixes #128 

It would be nice to see a new test to check that it works with other key encodings.

oed:
Added two tests, and caught a bug 👍

mirceanis (approved):
Looks great, Thanks!

PR #140, opened 2020-11-19
feat(jwt): add skewTime option that replaces NBF_SKEW if present

### LINKED ISSUE #136: JWT verification: skewed times comparison
Hi team!

I'm trying to verify some JWTs and I realised that `nbf`, `iat` and `exp` times are being compared to a skewed time (5 minutes) instead of the current one, is there any reason for that?

https://github.com/decentralized-identity/did-jwt/blob/c91d38cdd06635b418250048e329c509eab1e6d6/src/JWT.ts#L88
https://github.com/decentralized-identity/did-jwt/blob/7c6b10507a4c35c0bef0a82874507786b1dc8f56/src/JWT.ts#L252-L262

Thanks in advance!

javiesses:
This PR adds a new field called `skewTime` to `JWTVerifyOptions`. If it is present, it will be used to add/remove from `now` when performing timestamp validations when validating JWTs.

Closes #136

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -22,6 +22,7 @@ export interface JWTVerifyOptions {
   audience?: string
   callbackUrl?: string
   resolver?: Resolvable
+  skewTime?: number
 }
 
 export interface DIDAuthenticator {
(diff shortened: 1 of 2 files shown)


mirceanis (approved):
Fantastic!
Thank you for your contribution

PR #143, opened 2020-12-09
feat(did auth): update resolution of authentication entries in DIDDocument

mirceanis:
According to the did-core spec the `authentication` entries in a DID document can be either key ID `string` references or embedded `PublicKey` objects.

This PR adds support for that as well as preserving backwards compatibility with the `Authentication` data type used previously by some DID methods.


oed:
This is great, ty!

PR #149, opened 2021-02-09
feat: add `ES256KSigner` and `EdDSASigner` with uniform APIs

### LINKED ISSUE #78: Clarification on signer methods and return types

mirceanis:
fixes #78

I added `ES256KSigner` and `EdDSASigner` functions that create signer methods given a private key.
The private key encoding can be one of `Uint8Array`, or a string representing a `base64`, `base64url`, `base58btc`, or `hex` encoding.

This change set deprecates `SimpleSigner`, `NaclSigner`, and `EllipticSigner`.
They will be removed in the next major release.

This also expands the `Signer` interface to also accept `Uint8Arrays` for input.


src/__tests__/JWT-test.ts
```diff
@@ -1,7 +1,7 @@
 import { createJWT, verifyJWT, decodeJWT, createJWS, verifyJWS, resolveAuthenticator, NBF_SKEW } from '../JWT'
 import { TokenVerifier } from 'jsontokens'
-import SimpleSigner from '../SimpleSigner'
-import NaclSigner from '../NaclSigner'
+import SimpleSigner from '../signers/SimpleSigner'
```
rado0x54, line 3: Is moving the import for these deprecated Signer a breaking change or does Veramo export them globally once more? Otherwise I feel we might even drop the deprecated Implementations. :)

src/__tests__/SignerAlgorithm-test.ts
```diff
@@ -1,14 +1,14 @@
...
+import SimpleSigner from '../signers/SimpleSigner'
+import EllipticSigner from '../signers/EllipticSigner'
+import NaclSigner from '../signers/NaclSigner'
 import { ec as EC } from 'elliptic'
 import nacl from 'tweetnacl'
 import { base64ToBytes, base64urlToBytes, stringToBytes } from '../util'
 import { sha256 } from '../Digest'
 const secp256k1 = new EC('secp256k1')
-const privateKey = '278a5de700e29faae8e40e366ec5012b5ec63d36ec77e8a241154cc1d25383f'
+const privateKey = '278a5de700e29faae8e40e366ec5012b5ec63d36ec77e8a2417154cc1d25383f'
```
rado0x54, line 11: Really trying to understand what changed here & with the signature.

rado0x54: Is the change `a241154c` -> `a2417154c` intentional?

src/__tests__/SignerAlgorithm-test.ts
```diff
@@ -1,14 +1,14 @@
 import SignerAlgorithm from '../SignerAlgorithm'
 import { toSignatureObject } from '../VerifierAlgorithm'
-import SimpleSigner from '../SimpleSigner'
-import EllipticSigner from '../EllipticSigner'
-import NaclSigner from '../NaclSigner'
+import SimpleSigner from '../signers/SimpleSigner'
```
rado0x54, line 3: I would not expect any of these tests to change if these Signers were simple deprecated. What functional changes cause the tests changes here?

src/util.ts
```diff
@@ -10,12 +13,16 @@ export function bytesToBase64url(b: Uint8Array): string {
...
+ * @deprecated Please use base64ToBytes()
+ */
 export function base64urlToBytes(s: string): Uint8Array {
-  return u8a.fromString(s, 'base64url')
+  return base64ToBytes(s)
 }
 
 export function base64ToBytes(s: string): Uint8Array {
-  return u8a.fromString(s, 'base64pad')
+  const inputBase64Url = s.replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '')
```
rado0x54, line 24: Just wondering if regex vs substr `replace` has any efficiency differences. e.g. `s.replace(/\+/g, '-')` vs `s.replace('+', '-')`. Would not have expected regexes here :)

rado0x54 (changes requested):
I think the biggest question for me would be the changed tests around the deprecated signers. Other than that it looks good.

src/util.ts
```diff
@@ -59,6 +67,56 @@ export function toJose({ r, s, recoveryParam }: EcdsaSignature, recoverable?: bo
...
+ */
+export function parseKey(input: string | Uint8Array): Uint8Array {
+  if (typeof input === 'string') {
+    if (hexMatcher.test(input)) {
+      const parts = hexMatcher.exec(input)
+      return hexToBytes(parts[2])
+    } else if (base58Matcher.test(input)) {
+      return u8a.fromString(input, 'base58btc')
+    } else if (base64Matcher.test(input)) {
+      const inputBase64Url = input.replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '')
```
rado0x54, line 107: We can use `base64ToBytes` here, do deduplicate code, correct?

src/__tests__/SignerAlgorithm-test.ts
mirceanis: The private key "fix" forced all the signatures to change.
I can revert the key and use a leading zero.

src/util.ts
mirceanis: I initially tried the non-regex version but it only replaced the first matching character.

src/__tests__/SignerAlgorithm-test.ts
mirceanis: Yes, and thank you for raising this.

This key was not a proper hex encoding of 32 bytes.
The `SimpleSigner` used BigNumber when importing the privateKey which accepted the nonstandard encoding but generally the 32 bytes encoding in hex should be represented as 64 hex characters.

The reason I changed the key like this and why I didn't just add a leading zero is that this key was already used in other tests and could cause unnecessary confusion for others as it had for me.
It probably got copied wrong at some point.

I can revert this and add a leading zero to maintain the same signatures as a demonstration of equivalence.

src/__tests__/JWT-test.ts
mirceanis: It is not a breaking change since they are re-exported from `index.ts`

src/util.ts
mirceanis: yes!
great catch

src/__tests__/SignerAlgorithm-test.ts
rado0x54: ok, thanks

src/util.ts
rado0x54: ah ok, thanks

mirceanis:
@rado0x54
I pushed a couple of commits addressing the issues raised and I took some time to update the docs regarding signers.
Please take another look 🙏

src/util.ts
mirceanis: fixed

src/__tests__/SignerAlgorithm-test.ts
mirceanis: fixed.
I reverted the key(and added the leading zero) and the signatures.

src/__tests__/SignerAlgorithm-test.ts
mirceanis: fixed.
I reverted the key and added the leading zero.
Signatures went back to the original form.

rado0x54: Sounds good. I think it's totally fine to push that update in a separate PR, but in this context I was confused.

rado0x54 (approved):

awoie (approved):
LGTM. API looks so much better now!

oed:
This is great! Thanks :)

PR #156, opened 2021-03-23
feat: adapt to did core spec

### LINKED ISSUE #155: Adapt to latest did-core-spec
This library is using some assumptions that no longer hold true against the latest did core spec.
More specifically, did resolution should now use:
* `DidResolutionResult` instead of `DIDDocument`
* `verificationMethod` instead of `publicKey` in the `DIDDocument`
* and `blockchainAccountId` instead of `ethereumAddress`

mirceanis:
closes #155

#### added support for `blockchainAccountId`
This PR introduces verification capabilities for DID documents that use `blockchainAccountId` (with eip155)
Support for `ethereumAddress` is maintained for backward compatibility.

#### added support for `proofPurpose`
Verification can now be done using a `proofPurpose`, which filters the `verificationMethod` based on the entries in the respective section of the `DIDDocument`.

Examples:
```typescript
verifyJWT( token, { proofPurpose: 'authentication', ... })
verifyJWT( token, { proofPurpose: 'assertionMethod', ... })
verifyJWT( token, { proofPurpose: 'capabilityDelegation', ... })
// ...

// legacy API for 'authentication' purpose, maintained for backward compatibility.
verifyJWT( token, { auth: true, ... })`
```

For backward compatibility the verifier supports legacy DID resolvers that do not return `DIDResolutionResult` and also DID Documents that do not list `assertionMethod` at all.
This support will be removed once the ecosystem is more uniform with compliance.

#### testing
This PR also includes some hardening of existing tests. Using `expect.assertions(...)` to make sure that async tests are not falsely passing and replaced some test assertions with stricter versions and some `toMatchSnapshot` with `toEqual`


package.json
```diff
@@ -98,7 +74,8 @@
     "@stablelib/sha256": "^1.0.0",
     "@stablelib/x25519": "^1.0.0",
     "@stablelib/xchacha20poly1305": "^1.0.0",
-    "did-resolver": "^3.0.1",
+    "did-jwt": "^5.0.1",
```
oed, line 77: Why does it have itself as a dep? 😅

mirceanis: good catch!
this is still in progress. fixing it now..

mirceanis:
@oed now I think it's ready. please take another look

awoie (approved):
LGTM from a spec perspective although I think verificationRelationship is more appropriate than proofPurpose.

mirceanis:
> LGTM from a spec perspective although I think verificationRelationship is more appropriate than proofPurpose.

that's true, but it is so very verbose 😅

PR #159, opened 2021-03-24
fix: simplify expected Resolver type in verify methods

### LINKED ISSUE #158: the use of `Resolver` type from `did-resolver` is too restrictive
The `Resolver` class from `did-resolver` is declared with the private fields included, which makes it hard to implement matching type signatures.

For example, the following object cannot be used directly in `verifyJWT` or similar methods that require `Resolver`.
```typescript
const resolver = {
  resolve: (didURL: string, options: DIDResolutionOptions) => Promise.resolve(<DIDResolutionResult>{/*...*/})
}
```

The error is :
```
Type '{ resolve: (didUrl: string, options: DIDResolutionOptions) => Promise<DIDResolutionResult>; }' is missing the following properties from type 'Resolver': registry, cache
```
or
```
Type 'import("node_modules/did-resolver/lib/resolver").Resolver' is not assignable to type 'import("node_modules/did-jwt/node_modules/did-resolver/lib/resolver").Resolver'.
      Types have separate declarations of a private property 'registry'.
```

There is a workaround to use something like `(resolver as unknown) as Resolver` but this is not intuitive and annoying.

Only the required signature of the Resolver type should be required by this lib; without the private members.

mirceanis:
fixes #158
fixes https://github.com/decentralized-identity/did-jwt-vc/pull/68

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -16,12 +16,14 @@ export interface JWTOptions {
   expiresIn?: number
 }
 
+export type Public<T> = { [P in keyof T]: T[P] }
+
 export interface JWTVerifyOptions {
   /** @deprecated Please use `proofPurpose: 'authentication' instead` */
   auth?: boolean
   audience?: string
   callbackUrl?: string
-  resolver?: Resolver
+  resolver?: Public<Resolver>
   skewTime?: number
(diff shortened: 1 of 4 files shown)


simonas-notcat (approved):
LGTM

PR #161, opened 2021-04-30
Use stable stringify when creating payload segments

zbarbuto:
When creating the segments of a JWT, regular `JSON.stringify()` was being used. Because the ordering is not guaranteed with JSON.stringify - this makes re-creating the JWT out of a W3C format credential very difficult.

By using [json-stable-stringify](https://www.npmjs.com/package/json-stable-stringify) there can be a guaranteed ordering of the properties, meaning as long as the payload contains exactly the same content, you will always get the same JWT (see added test which fails without these changes)

I have also replaced ncal-did with regular jwt decode in the test - nacl-did is [marked as deprecated](https://github.com/uport-project/nacl-did) and was only used in this one place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "jsontokens": "3.0.0",
     "microbundle": "0.13.0",
     "mockdate": "3.0.5",
-    "nacl-did": "1.0.1",
     "prettier": "2.2.1",
     "regenerator-runtime": "0.13.7",
     "semantic-release": "17.4.2",
@@ -88,6 +87,7 @@
     "did-resolver": "^3.1.0",
     "elliptic": "^6.5.4",
     "js-sha3": "^0.8.0",
+    "json-stable-stringify": "^1.0.1",
     "uint8arrays": "^2.1.3"
   },
   "standard": {
--- a/package.json
+++ b/package.json
@@ -88,10 +88,10 @@
     "@stablelib/sha256": "^1.0.0",
     "@stablelib/x25519": "^1.0.0",
     "@stablelib/xchacha20poly1305": "^1.0.0",
+    "canonicalize": "^1.0.5",
     "did-resolver": "^3.1.0",
     "elliptic": "^6.5.4",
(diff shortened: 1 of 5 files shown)


mirceanis (changes requested):
Hi, thank you for the contribution!

It looks like you are looking for a way to canonicalize the payload before signing.
I think if you are working with already canonical JSON, then `JSON.stringify` will behave properly.
I'm not against performing this canonicalization, but I think that should be an opt-in, not a default. Existing users may not like their payloads to be modified.

That being said, the standard way of doing canonicalization of JSON in cryptographic situations is JCS, described by [RFC8785](https://tools.ietf.org/html/rfc8785) which also lists [canonicalize](https://github.com/erdtman/canonicalize) as a zero-dependency javascript implementation of JCS.

I would suggest using `canonicalize` instead of `json-stable-stringify` since the later also seems to have [serious issues](https://github.com/substack/json-stable-stringify/issues) and doesn't look like it's maintained.

Also, please make a separate PR for the removal of `nacl-did` from the `devDependencies` and from tests.

mirceanis:
@zbarbuto It seems that other projects are successfully using [fast-json-stable-stringify](https://github.com/epoberezkin/fast-json-stable-stringify) to stabilise the payload [before sending it to be signed](https://github.com/ceramicnetwork/key-did-provider-ed25519/blob/07ce5976c22a0798416ef2c1217c86f8e99720fb/src/index.ts#L19)

You might want to try that one and see if it resolves your issue.
If that works, please close this PR, otherwise please make the required changes.
Thank you for your contribution!

zbarbuto:
As requested I have replaced `json-stable-stringify` with `canonicalize` and made canonicalization optional with a new `canonicalize` option in `createJWT` / `createJWS`

zbarbuto:
CI failure does not appear related to my changes that I can see - the issue is in uploading coverage reports.

mirceanis (approved):
looks great, thanks for the contribution!

PR #162, opened 2021-05-02
Add support for ECDH-1PU+XC20PKW

awoie:
Add support for ECDH-1PU+XC20PKW


src/Digest.ts
```diff
@@ -27,14 +27,15 @@ const lengthAndInput = (input: Uint8Array): Uint8Array => u8a.concat([writeUint3
...
 // https://github.com/panva/jose/blob/master/lib/jwa/ecdh/derive.js
-export function concatKDF(secret: Uint8Array, keyLen: number, alg: string): Uint8Array {
+export function concatKDF(secret: Uint8Array, keyLen: number, alg: string, apu?:string, apv?:string): Uint8Array {
   if (keyLen !== 256) throw new Error(`Unsupported key length: ${keyLen}`)
   const value = u8a.concat([
     lengthAndInput(u8a.fromString(alg)),
-    lengthAndInput(new Uint8Array(0)), // apu
-    lengthAndInput(new Uint8Array(0)), // apv
+    lengthAndInput((apu == undefined) ? new Uint8Array(0) : u8a.fromString(apu)),
+    lengthAndInput((apv == undefined) ? new Uint8Array(0) : u8a.fromString(apv)),
```
oed, line 35: This is supposed to be `utf8` to bytes?

src/JWE.ts
```diff
@@ -116,7 +118,7 @@ export async function decryptJWE(jwe: JWE, decrypter: Decrypter): Promise<Uint8A
   } else if (!jwe.recipients || jwe.recipients.length === 0) {
     throw new Error('Invalid JWE')
   } else {
-    for (let i = 0; !cleartext && i < jwe.recipients.length; i++) {
+    for (let i = 0; !cleartext && i < jwe.recipients.length; i++) {
```
oed, line 121: remove trailing spaces I guess?

src/__tests__/jwe-vectors.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable */
 // it.skip('not a test', () => {})
 module.exports = {
-  dir: {
+  ecdhEsXc20p: {
```
oed, line 4: This is not correct. there is no key exchange involved when using direct encryption!
The protected header of the vector below states:
```json
{
  "alg": "dir",
  "enc": "XC20P"
}
```

src/__tests__/jwe-vectors.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
       },
     ]
   },
-  x25519: {
+  ecdhEsXc20Pkw: {
```
oed, line 48: This is correct however.

src/__tests__/JWE.test.ts
```diff
@@ -1,58 +1,87 @@
...
+import { bytesToBase64, decodeBase64url, encodeBase64url } from '../util'
 import * as u8a from 'uint8arrays'
 import { randomBytes } from '@stablelib/random'
 import { generateKeyPairFromSeed } from '@stablelib/x25519'
 
 describe('JWE', () => {
   describe('decryptJWE', () => {
-    describe('Direct encryption', () => {
-      test.each(vectors.dir.pass)('decrypts valid jwe', async ({ key, cleartext, jwe }) => {
+    describe('ECDH-ES (X25519), Direct Mode with XC20P content encryption', () => {
```
oed, line 12: Change back.

src/__tests__/JWE.test.ts
```diff
@@ -186,4 +215,112 @@ describe('JWE', () => {
...
+
+      it('Creates with only ciphertext', async () => {
+        expect.assertions(3)
+        const jwe = await createJWE(cleartext, [encrypter])
+        expect(jwe.aad).toBeUndefined()
+        expect(JSON.parse(decodeBase64url(jwe.protected))).toEqual({ enc: 'XC20P', skid: skid })
+        expect(await decryptJWE(jwe, decrypter)).toEqual(cleartext)
+      })
+
+      it('Creates with sender kid', async () => {
```
oed, line 241: This test looks identical to the test above it.

oed (changes requested):
Looks great! Just requesting some minor changes

src/__tests__/jwe-vectors.js
awoie: Right, will rename it to xc20p though because it is jwe-vector.js which i guess can expand on other content encryption algorithm than XC20p. IMO, it is nicer to group the test vectors by algorithm.

src/__tests__/JWE.test.ts
awoie: yep, same as above. Will change it to `Direct Encryption with XC20P`

src/Digest.ts
awoie: I will change the parameters for apu and apv to producer/consumerInfo as UInt8Arrays. 

The JWE spec defines what values and encodings are expected. In case of 1PU, apu and apv supposed to be the base64url decoded apu/apv which in 1PU's case is always some UTF-8 bytes but to be more general, I will change that.

src/__tests__/jwe-vectors.js
oed: makes sense 👍

src/__tests__/JWE.test.ts
```diff
@@ -186,4 +215,152 @@ describe('JWE', () => {
...
+      it('Creates with skid, kid, no apu and no apv', async () => {
+        const kid = 'did:example:receiver#key-1'
+        const skid = 'did:example:sender#key-1'
+        const encrypter = x25519AuthEncrypter(recipientKey.publicKey, senderKey.secretKey, { kid, skid } )
+        expect.assertions(6)
+        const jwe = await createJWE(cleartext, [encrypter])
+        expect(jwe.aad).toBeUndefined()
+        expect(JSON.parse(decodeBase64url(jwe.protected))).toEqual({ enc: 'XC20P', skid: skid })
+        expect(jwe.recipients[0].header.kid).toEqual(kid)
+        expect(jwe.recipients[0].header.apu).toEqual(encodeBase64url(skid))
```
oed, line 248: How would this work with an encrypted skid?

src/xc20pEncryption.ts
```diff
@@ -78,6 +79,84 @@ export function x25519Encrypter(publicKey: Uint8Array, kid?: string): Encrypter
...
+
+  const alg = 'ECDH-1PU+XC20PKW'
+  const keyLen = 256
+  const crv = 'X25519'
+
+  // It is RECOMMENDED by the ECDH-1PU spec to set apu and apv 
+  // to base64url encoded kid and base64url encoded skid in the
+  // recipient header. For the KDF, if no apu and apv is provided,
+  // it is RECOMMENDED that PartyVInfo and PartyUInfo contain
+  // the base64url decoded skid and kid.
```
oed, line 100:
```suggestion
  // the base64url encoded skid and kid.
```
Should be this?

src/xc20pEncryption.ts
```diff
@@ -78,6 +79,84 @@ export function x25519Encrypter(publicKey: Uint8Array, kid?: string): Encrypter
...
+  function setPartyInfo(partyInfo, fallback): { encoded: string, raw: Uint8Array } {
+    if (typeof partyInfo === 'undefined') {
+      return (typeof fallback !== 'undefined') ? 
+        { encoded: encodeBase64url(fallback), raw: fromString(fallback) } :
+         { encoded: undefined, raw: undefined }
```
oed, line 105: Plz make this more readable by using another `if` statement.

src/xc20pEncryption.ts
```diff
@@ -78,6 +79,84 @@ export function x25519Encrypter(publicKey: Uint8Array, kid?: string): Encrypter
...
+      header: {
+        alg,
+        iv: bytesToBase64url(res.iv),
+        tag: bytesToBase64url(res.tag),
+        epk: { kty: 'OKP', crv, x: bytesToBase64url(epk.publicKey) }
+      }
+    }
+    if (typeof options.kid !== 'undefined') recipient.header.kid = options.kid
+    if (typeof partyUInfo.encoded !== 'undefined') recipient.header.apu = partyUInfo.encoded
+    if (typeof partyVInfo.encoded !== 'undefined') recipient.header.apv = partyVInfo.encoded
```
oed, line 142: Why not just this?
```suggestion
    if (options.kid) recipient.header.kid = options.kid
    if (partyUInfo.encoded) recipient.header.apu = partyUInfo.encoded
    if (partyVInfo.encoded) recipient.header.apv = partyVInfo.encoded
```

oed (changes requested):
Lgtm, just a few minor tweaks needed.

src/__tests__/JWE.test.ts
awoie: `skid` as per ECDH-1PU is just an OPTIONAL `kid` of the sender in the protected header. The discussion in the DIDComm WG will likely conclude with either including a `encrypted_skid` (or `encrypted_spk`) in the recipient header. So, this is unrelated. In this implementation `skid` is just a blob without any meaning similar to `kid`.

Adding `encrypted_skid` (or `encrypted_spk`) could be subject to another PR. Would you agree?

src/xc20pEncryption.ts
awoie: Nope, it is `decoded`. It means, you should decodeBase64url(skid) and decodeBase64url(kid) for the KDF. However, `apu` and `apv` will contain the base64url encoded values.

awoie: But it raises the question if it is a good idea to expect `apu` and `apv` to be the plain string values rather base64url encoded values in the high level x25519AuthEncrypter API:

```
export function x25519AuthEncrypter(recipientPublicKey: Uint8Array, senderSecretKey: Uint8Array, 
  options: Partial<X25519AuthEncryptParams> = {}): Encrypter
```

I thought it might be more easy for a developer that x25519AuthEncrypter ensure base64url encoding. Thoughts?

src/Digest.ts
```diff
@@ -27,14 +28,15 @@ const lengthAndInput = (input: Uint8Array): Uint8Array => u8a.concat([writeUint3
...
-export function concatKDF(secret: Uint8Array, keyLen: number, alg: string): Uint8Array {
+export function concatKDF(secret: Uint8Array, keyLen: number, alg: string, producerInfo?:Uint8Array, consumerInfo?:Uint8Array): Uint8Array {
   if (keyLen !== 256) throw new Error(`Unsupported key length: ${keyLen}`)
   const value = u8a.concat([
-    lengthAndInput(u8a.fromString(alg)),
-    lengthAndInput(new Uint8Array(0)), // apu
-    lengthAndInput(new Uint8Array(0)), // apv
+    lengthAndInput(u8a.fromString(alg)),    
+    lengthAndInput(typeof producerInfo === 'undefined' ? new Uint8Array(0) : producerInfo),
+    lengthAndInput(typeof consumerInfo === 'undefined' ? new Uint8Array(0) : consumerInfo),
```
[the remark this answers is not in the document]
awoie: done

src/__tests__/JWE.test.ts
oed: Sure, was just thinking that passing an object here (JWE) rather than a string will likely break the current implementation.

src/xc20pEncryption.ts
oed: Don't have a strong sense of how developers  would want to use apu / apv.

awoie: apu and apv are now expected to be base64url encoded since technically, it could also contain binary data

src/xc20pEncryption.ts
awoie: this is no longer needed, removed it entirely

.circleci/config.yml
```diff
@@ -61,7 +61,7 @@ jobs:
   verify-unit-tests:
     working_directory: ~/project
     docker:
-      - image: circleci/node:12
+      - image: circleci/node:15
```
oed, line 64: Can we please use 14 here instead since that's LTS.

oed (commented):
@awoie have a hard time reviewing the new changes since it looks like the git history got screwed up.

awoie:
Closing this PR since we have a clean new one that only targets 1PU.

PR #166, opened 2021-05-10
Support multiple keyAgreement keys

dankelleher:
If a DID contains more than one valid keyAgreement key, payloads should be encrypted for both of them, ensuring that all keyholders can decrypt.


rado0x54:
This matches the behavior described in the DIDComm Spec for message encryption: https://identity.foundation/didcomm-messaging/spec/#did-document-keys.

Therefore I vote to accept this PR. It should be fully backwards compatible from what I can see.

oed (approved):
lgtm

PR #179, opened 2021-06-01
chore(ci): add github workflows

mirceanis:
A maintenance PR that enables the use of github actions for building / testing / releasing.

This is a prerequisite to removing the dependency on circleci.
The results of this PR are already visible in the `Build and Test NODE / build-test (pull_request)` status check which runs the `.github/workflows/test.yml` workflow


awoie (approved):
lgtm

PR #186, opened 2021-06-08
feat: enable remote ECDH for JWE [de]encrypters

### LINKED ISSUE #183: add ability to use external `sharedSecret` function
### Context
When creating auth JWE or decrypting JWE, the current assumption is that there is access to secretKey material so that `sharedKey()` can be called directly.
Examples [here](https://github.com/decentralized-identity/did-jwt/blob/0db32116647bd69d265eafbee3094ffde1df937c/src/xc20pEncryption.ts#L275), [here](https://github.com/decentralized-identity/did-jwt/blob/0db32116647bd69d265eafbee3094ffde1df937c/src/xc20pEncryption.ts#L179) and [here](https://github.com/decentralized-identity/did-jwt/blob/0db32116647bd69d265eafbee3094ffde1df937c/src/xc20pEncryption.ts#L313)

### Problem
This prevents external crypto implementations from being used along with this library.

### Solution
In a similar fashion to the [`Signer`](https://github.com/decentralized-identity/did-jwt/blob/0db32116647bd69d265eafbee3094ffde1df937c/src/JWT.ts#L7) pattern, I'm proposing that a similar interface be used to allow the use of external crypto.

### Example:
```typescript
type ECDH = (theirPublicKey: Uint8Array) => Promise<Uint8Array>
```
with a possible implementation along the lines of:
```typescript
export function createX25519ECDH(mySecretKey: Uint8Array): ECDH {
  return async (theirPublicKey: Uint8Array): Promise<Uint8Array> => {
    return sharedKey(mySecretKey, theirPublicKey)
  }
}
```

mirceanis:
With this change, users will be able to use the JWE functionality even without direct access to sender or recipient `secretKey` material, as long as they can provide a wrapper method that resolves to the shared secret.

This also adds a `createX25519ECDH(mySecretKey: Uint8Array): ECDH` method that creates such a wrapper from an existing key.

The signature of the wrapper is this:
```typescript
type ECDH = (theirPublicKey: Uint8Array) => Promise<Uint8Array>
```

fixes #183

### DIFF
--- a/src/ECDH.ts
+++ b/src/ECDH.ts
@@ -2,15 +2,21 @@ import { sharedKey } from '@stablelib/x25519'
 
 /**
  * A wrapper around `mySecretKey` that can compute a shared secret using `theirPublicKey`
- * The promise should resolve to a Uint8Array containing the raw shared secret.
+ * The promise should resolve to a `Uint8Array` containing the raw shared secret.
  *
+ * This method is meant to be used when direct access to a secretKey is impossible or not desired.
+ *
+ * @param theirPublicKey `Uint8Array` the other party's publicKey
+ * @returns a `Promise` that resolves to a `Uint8Array` representing the computed shared secret
  */
 export type ECDH = (theirPublicKey: Uint8Array) => Promise<Uint8Array>
 
 /**
  * Wraps an X25519 secretKey into an ECDH method that can be used to compute a shared secret with a publicKey.
- * @param mySecretKey A `Uint8Array` representing the bytes of my secret key
+ * @param mySecretKey A `Uint8Array` of length 32 representing the bytes of my secret key
  * @returns an `ECDH` method with the signature `(theirPublicKey: Uint8Array) => Promise<Uint8Array>`
+ *
+ * @throws 'invalid_argument:...' if the secret key size is wrong
  */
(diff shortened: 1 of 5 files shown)


src/ECDH.ts
```diff
@@ -0,0 +1,31 @@
+import { sharedKey } from '@stablelib/x25519'
+
+/**
+ * A wrapper around `mySecretKey` that can compute a shared secret using `theirPublicKey`
```
awoie, line 4:
```suggestion
 * A wrapper around `mySecretKey` that can compute a shared secret using `theirPublicKey`.
```

src/xc20pEncryption.ts
```diff
@@ -5,70 +5,126 @@ import { concatKDF } from './Digest'
...
+ * Extra header parameters for JWE using authenticated encryption
+ */
 export type AuthEncryptParams = {
+  /**
+   * recipient key ID
+   */
   kid?: string
+
+  /**
+   * sender key ID
```
awoie, line 20:
```suggestion
   * See {@link https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-03#section-2.1.1}
```

awoie (approved):
lgtm

PR #188, opened 2021-06-11
fix: removed skid from the recipient header

awoie:
skid is no longer (was meant to be encrypted to recipient) in the recipient header and only added to the protected header.


mirceanis (approved):
looks good

PR #189, opened 2021-06-24
fix: add better error messages

mirceanis:
This PR is part of a greater effort to reduce the maintenance burden by using a standard set of scripts and tools on all repositories currently in maintenance.

Also in this PR:
* update dependencies
* reformatting code
* apply same linter and compiler settings as other libs
* Export some extra types related to JWE processing


jasonhealy (approved):
Looks good!

PR #200, opened 2021-09-28
feat: support `publicKeyMultibase`

yhuard:
This PR adds support for [publicKeyMultibase](https://www.w3.org/TR/did-core/#dfn-publickeymultibase).

TBD: right now, it decodes `base16`, `base58btc`, `base64`, or `base64url` multibase strings only. We could add support for other bases too: `base2`, `base8`, `base10`, `base32`, `base36`, and all their variants. The [multiformats library](https://github.com/multiformats/js-multiformats) doesn't provide a universal decoder. That's why we have to compose our own decoder like this:

```js
 const baseDecoder = base16.decoder.or(base58btc.decoder.or(base64.decoder.or(base64url.decoder)))
```

If you think that `base16`, `base58btc`, `base64`, and `base64url` are enough, then this PR is ready.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,9 +84,10 @@
     "@stablelib/x25519": "^1.0.1",
     "@stablelib/xchacha20poly1305": "^1.0.1",
     "canonicalize": "^1.0.5",
-    "did-resolver": "^3.1.0",
+    "did-resolver": "^3.1.1",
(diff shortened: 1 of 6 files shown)


mirceanis (approved):
This is great!
Thank you for your contribution.

PR #201, opened 2021-10-11
feat: resolve DID document from self-signed JWT

yhuard:
Hi!

If the `iss` contains https://self-issued.me (covers both https://identity.foundation/did-siop/ and https://openid.net/specs/openid-connect-self-issued-v2-1_0.html), then the DID must be extracted from another property. 
In the case of SIOP DID Profile v0.1, a `did` property was required in the payload. In the new SIOP v2 specs, I haven't seen any reference to this `did` property, so I guess it's been deprecated. If `did` is not present, then the library will check the `kid` header. I'm not sure if the DID can come from other properties as well. It seems that the `sub` could also contain the DID:

> The sub (subject) Claim value is either the base64url encoded representation of the thumbprint of the key in the sub_jwk Claim or a decentralized identifier.

— https://openid.net/specs/openid-connect-self-issued-v2-1_0.html#section-9

In any case, the change I propose is fairly simple. If the `iss` contains `https://self-issued.me`, then the lib checks first if there's a `did` prop. If not, it checks the `kid` header. If the `iss` doesn't contain `https://self-issued.me`, the lib will try to use the `iss` prop as DID. If the DID is empty or undefined, the lib throws an error. It's more or less the same logic that Sphereon implemented https://github.com/Sphereon-Opensource/did-auth-siop/blob/develop/src/did-jwt-fork/JWT.ts#L330

Regarding the tests, I've made some extra fixes. Before, the mocked resolver would always return the same DID document, no matter what DID was being resolved. Now, it returns 2 different DID Documents depending on which DID is being resolved (`aud` or `did`). I have also updated the pregenerated JWT to use the existing DID `did` instead of `did:ethr:0x90e45d75bd1246e0924872018647dba996a8e7b9` (which should not resolve to anything).

Please let me know if changes are needed.

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -133,7 +133,8 @@ export const SUPPORTED_PUBLIC_KEY_TYPES: PublicKeyTypes = {
   EdDSA: ['ED25519SignatureVerification', 'Ed25519VerificationKey2018'],
 }
 
-export const SELF_ISSUED = 'https://self-issued.me'
+export const SELF_ISSUED_V2 = 'https://self-issued.me/v2'
+export const SELF_ISSUED_V0_1 = 'https://self-issued.me'
 
 type LegacyVerificationMethod = { publicKey?: string }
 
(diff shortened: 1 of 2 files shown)


nklomp:
> The current specs https://openid.net/specs/openid-connect-self-issued-v2-1_0.html are contradictory when it comes to `sub_jwk`:

Hehe, indeed. The spec is contradictory in many ways (mixing plurals and singulars all over the place and then having a singular object in the example, where a key is plural etc). I will be reaching out to the editors about that anyway. 

Also thanks for the clarification of the sub_jwk. I guess we can strip some code from our solution as we did create the sub_jwk according to the 2nd options ;).

PR looks good to me. Thanks for the work! :)

yhuard:
@mirceanis can you review it when you have time please? :)

PR #207, opened 2021-12-03
fix: add missing quotes around undefined

yhuard:
This is a quick fix adding missing quotes around `undefined`.

Without the quotes, `typeof blockchainAccountId !== undefined` always return `true`, and all the `authenticators` are added to `blockchainAddressKeys`, which causes the lib to unnecessarily call `signer = verifyRecoverableES256K(data, signature, blockchainAddressKeys)` (line 78) when `signer` is falsy.


mirceanis (approved):
nice find, thanks for spotting and fixing that!

PR #211, opened 2022-01-10
chore: Finalize transition to ES modules

ukstv:
Current `did-jwt` package exposes both CJS and ESM modules. It seems to work, for most of the cases. Really, support for ESM is partial, and relies on non-strict Node.js behaviour with regards to module handling. This PR makes it right.

Full support for ESM means exposing ESM js file when imported from ESM js file. One could see that currently this results in importing CJS file, which misses the point of having CJS and ESM files in the package.

**To see that CJS is all right:** Run Node.js REPL inside the did-jwt repository, use command `require.resolve('did-jwt')`. This should result in `/<full-path-from-root>/did-jwt/lib/index.js`, which indicates that CJS is exposed all right.

**To see that ESM is (not) all right:** Create an empty npm package via `npm init`. Add a dependency: `npm add did-jwt`. Make this package ESM: add a line `"type": "module"` to `package.json`. Then create a file `index.js` with the following contents:
```
import.meta.resolve('did-jwt').then(path => console.log(path))
```
Run it as `node --experimental-import-meta-resolve ./foo.js`. `--experimental-import-meta-resolve` enables `import.meta.resolve`, so that we could see what file is going to be actually imported.

With the upstream, the result is `/<full-path-from-root>/did-jwt/lib/index.js`, which is CJS file, not ESM, as expected. This PR makes it correct: import from ESM module (as witnessed by `import.meta.resolve`) results in `/<full-path-from-root>/did-jwt/lib/index.module.js`.


mirceanis (approved):
looks great, thank you for the contribution!

Issue #170, opened 2021-05-18
Rename this repo to "did-jose"


oed:
This package now contains many things related to the JOSE spec. We should modularize it to a few different (to be determined) packages.

Suggestion:
* `@did-jose/jws`
* `@did-jose/jwe`
* `@did-jose/jwt`
* `@did-jose/signing-*`
* `@did-jose/encryption-*`

peacekeeper:
Hey, since this is a work item of the I&D WG, if you rename it then maybe you could also update the entry in the [list of Work Items](https://github.com/decentralized-identity/identifiers-discovery/#work-items), and while doing that also check the work item leads that are listed there?

Perhaps we will also ask you at some point to give a general update on the work item progress during one of the weekly I&D WG calls.

TomCJones:
I am unclear where this label is applied. Jose has a header.  Some of what is proposed is already in that header. Perhaps all of it should be in the Jose header. Otherwise difference levels of the code would be involved. That seems wrong.

oed:
@TomCJones Are you replying to the wrong issue? :)

awoie:
I agree that renaming/repackaging makes sense but I don't have a strong stance here.

oed:
@mirceanis Would love to hear your thoughts here!

mirceanis:
Modularising this repo sounds like a good idea.
I'm not yet sure about the actual module layout.
I think the goals should be to reduce lib size (and transitive dependencies) for users that only use a subset of modules, and to allow extension of the supported algorithms using user-provided modules.

oed:
> I think the goals should be to reduce lib size (and transitive dependencies) for users that only use a subset of modules, and to allow extension of the supported algorithms using user-provided modules.

Yep agreed! That was what I was aiming for above, e.g. with `@did-jose/signing-*` representing a specific alg.

oed:
This is not stale.

Issue #146, opened 2021-01-17
Warn users to avoid using unregistered ES256K-R alg


OR13:
Either register it here: https://www.iana.org/assignments/jose/jose.xhtml

Or warn users not to use it.... it should not be listed next to registered algorithms without making it clear that it was invented by consensys and never registered.

mirceanis:
fixed in #149 with a note in the readme

PR #212, opened 2022-01-17
Developing a Signer for P-256. (work in progress)

bshambaugh:
This may not be complete as is.

I am in the process of implementing a signer for a did:key provider for use with ceramic. it is like the lines below, which utilize this library.

https://github.com/ceramicnetwork/key-did-provider-ed25519/blob/main/src/index.ts#L53
https://github.com/ceramicnetwork/key-did-provider-secp256k1/blob/master/src/index.ts#L56

I will add to this PR as I visit JWT.(test).ts, VerifierAlgorithm.(test).ts, and SignerAlgorithm.(test).ts in the process of implementing https://eips.ethereum.org/EIPS/eip-2844 for the P-256 (secp256r1) key.

@oed has been reviewing a resolver that includes P-256 for ceramic https://github.com/ceramicnetwork/js-ceramic/pull/1884 , so this is a heads up to him that I am in the process of working on this code base.

I am still not sure about the acronym I should use for the type:
Secp256r1VerificationKey2022 (<curve-specific><purpose>Key<date> like: https://lists.w3.org/Archives/Public/public-credentials/2018Mar/0104.html) 
or P256Key2021(https://github.com/transmute-industries/did-key.js/blob/c9a478da12508fefd8018d82ab95638ea6980dd7/packages/did-key-test-vectors/src/secp256r1/did-key-secp256r1-case-1.json#L81)
sounds like a reasonable acronym to use in JWT.ts, however I only support application/did+json at the moment, not application/did+ld+json which seems to prefer publicKeyBase58 and an acronym like P256Key2021. 

Thus, the less descriptive "JsonWebKey2020" seems like the way to go.

Note. ES256 and P-256 are mentioned here: https://www.iana.org/assignments/jose/jose.xhtml


mirceanis:
Excellent start!

I agree that `JsonWebKey2020` seems like a good choice.

bshambaugh:
[mirceanis](https://github.com/mirceanis) , I think I am at a state where it looks a little better. I'll announce this to the CCG list and to the people with 3box/ceramic

bshambaugh:
The verifierAlg may note be set up to work with cosmos. I'm looking into this. 

See: https://github.com/bshambaugh/did-jwt/blob/master/src/__tests__/VerifierAlg/ES256VerifierAlg.test.ts#L49
and https://github.com/bshambaugh/did-jwt/blob/master/src/blockchains/cosmos.ts#L8

bshambaugh:
I also publicly mentioned this PR here: https://lists.w3.org/Archives/Public/public-credentials/2022Mar/0023.html

src/JWT.ts
```diff
@@ -91,6 +91,8 @@ export interface PublicKeyTypes {
   [name: string]: string[]
 }
 export const SUPPORTED_PUBLIC_KEY_TYPES: PublicKeyTypes = {
+  ES256: ['JsonWebKey2020', 'EcdsaSecp256r1VerificationKey2022'],
+  'ES256-R': ['JsonWebKey2020', 'EcdsaSecp256r1VerificationKey2022'],
```
oed, line 95: Is `ES256-R` really needed? Afaik no one uses this?

package.json
```diff
@@ -89,7 +91,8 @@
     "elliptic": "^6.5.4",
     "js-sha3": "^0.8.0",
     "multiformats": "^9.4.10",
-    "uint8arrays": "^3.0.0"
+    "uint8arrays": "^3.0.0",
+    "yarn": "^1.22.17"
```
oed, line 95: yarn should not be a dependency here!

oed (commented):
Don't have time to look though all of this right now. Just some quick feedback from a glance.

package.json
bshambaugh: Okay, I took this out. Thanks.

bshambaugh:
Re-posting here from the Ceramic Discord #didathing channel:
" I was thinking as I develop a key-did-provider and add did-jwt-vc I would want P-256 support (https://github.com/decentralized-identity/did-jwt/pull/212). I went through the code for did providers for ed25519 and secp256k1 and tried to infer what to add in terms of code and tests with the goal of eventually supporting verifiable credentials. I just started making noises about did-jwt work after working mostly silently for about a month an a half. I hope that the code is of decent quality. I think how everything works will become clear when I build key-did-provider-p256 and add in my remote signer. ( I tried to imply above that I could probably sandbox things locally for awhile). 

My mind goes everywhere. Sometimes I imagine walking someone through what I did, or at least being open to answer questions as I reorganized the library, I know it is not perfect, and I know there was an issue in did-jwt that talked about breaking up did-jwt (https://github.com/decentralized-identity/did-jwt/issues/170, and I did not do that, but perhaps breaking up the code by signer was a start.. Sometimes I imagine someone telling me, oh, yeah that wasn't the whole purpose of the library when Pelle and the rest of the uPort team started off with it, and it still isn't for anyone, and your addition doesn't make any sense. It has been a short time. People know about it on the CCG list and here. The API of the library, how people use it with other code, wasn't changed, but I felt the need to reorganize it. Eventually I imagine this reorganization will  / or might help with P-384 and P-521 & RSA and ?? (but that isn't a priority yet...but could help with support for these curves (maybe in line with CCG initiatives for legacy HSM as mentioned by Manu Sporny).
"
"Basically, if it sees ES256 as a signer, it knows what to do. I think I need JWTs, especially when did-jwt-vc is included. I'll try building more of blinky / LoRa thing out and see what happens (maybe this is more critical in a month). did-jwt seems like a core library called by many repositories, so I hope anything I do is clean and makes sense for future people."
"I may not need the organization suggested in issue 170 just yet . I broke stuff up because otherwise the files would be massive and functionality and tests would be squashed in one place making following the code very confusing.
Why have ES256 and ES256K code together in one file if it can be helped?
I basically copied a lot of secp256k1 / ES256K code and reformulated it for P-256/ES256 and broke out shared functionality and put that in its own place.  The new code for P-256/ES256 was very similar as the curves and signing alg are very similar. I assumed logic for eip155, bip122,  caip-10, (and now cosmos) will just transfer over. It's basically a clone. Swap out kobiltz for prime in the curve. "

src/JWT.ts
bshambaugh: It clearly seemed like a non-standard key type. Adding recoverySigner , was a "oh why not" kind of thing. But 'ES256-R' and its earlier 'ES256-K' were use at your own discretion kind of things.

bshambaugh: reference: https://github.com/decentralized-identity/did-jwt/issues/146

OR13:
I strongly suggest you comment on https://github.com/w3c/vc-wg-charter/issues/82.

src/JWT.ts
mirceanis: I agree with @oed, `ES256-R` is not needed.
`ES256K-R` existed as a pattern and was in use before it got a name.
It exists now to allow for `blockchainAccountId` verification methods to be used as verifiers.
We could probably phase it out of existence if we accept 1 bit less of security for ES256K + blockchainAccountId

src/SignerAlgorithm.ts
```diff
@@ -1,49 +1,25 @@
...
-      return toJose(signature, recoverable)
-    } else {
-      if (recoverable && typeof fromJose(signature).recoveryParam === 'undefined') {
-        throw new Error(`not_supported: ES256K-R not supported when signer doesn't provide a recovery param`)
-      }
-      return signature
-    }
-  }
-}
+import * as ES256KSignerAlg from './SignerAlg/ES256KSignerAlg'
```
mirceanis, line 3: These imports can be rephrased as:
```suggestion
import { ES256KSignerAlg } from './SignerAlg/ES256KSignerAlg'
```

and then the code below is simplified as well

src/SignerAlgorithm.ts
```diff
@@ -1,49 +1,25 @@
...
 
 interface SignerAlgorithms {
   [alg: string]: SignerAlgorithm
 }
 
 const algorithms: SignerAlgorithms = {
-  ES256K: ES256KSignerAlg(),
+  ES256: ES256SignerAlg.ES256SignerAlg(),
+  'ES256-R': ES256SignerAlg.ES256SignerAlg(true),
+  ES256K: ES256KSignerAlg.ES256KSignerAlg(),
```
mirceanis, line 15: With the above suggestion, this would simplify to:
```suggestion
  ES256K: ES256KSignerAlg(),
```

src/VerifierAlg/ES256VerifierAlg.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  if (!signer && blockchainAddressKeys.length > 0) {
+    signer = verifyRecoverableES256(data, signature, blockchainAddressKeys)
+  }
+
+  if (!signer) throw new Error('invalid_signature: Signature invalid for JWT')
+  return signer
+}
+
+export function verifyRecoverableES256(
```
mirceanis, line 43: This is probably not needed since ES256-R is not needed

src/VerifierAlgorithm.ts
```diff
@@ -1,12 +1,10 @@
...
-import { sha256, toEthereumAddress } from './Digest'
-import { verify } from '@stablelib/ed25519'
 import type { VerificationMethod } from 'did-resolver'
-import { bases } from 'multiformats/basics'
-import { hexToBytes, base58ToBytes, base64ToBytes, bytesToHex, EcdsaSignature, stringToBytes } from './util'
-import { verifyBlockchainAccountId } from './blockchains'
+import { base64ToBytes, bytesToHex, EcdsaSignature } from './util'
 
-const secp256k1 = new EC('secp256k1')
+import * as Ed25519VerifierAlg from './VerifierAlg/Ed25519VerifierAlg'
```
mirceanis, line 4: flagging this for simplification of imports:
```suggestion
import { Ed25519VerifierAlg } from './VerifierAlg/Ed25519VerifierAlg'
```

src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts
```diff
@@ -0,0 +1,205 @@
...
+  publicKeyBase58?: string
+  publicKeyBase64?: string
+  publicKeyJwk?: JsonWebKey
+  publicKeyHex?: string
+  publicKeyMultibase?: string
+  blockchainAccountId?: string
+  ethereumAddress?: string
+}
+
+interface VerificationMethodCommonSigner {
```
mirceanis, line 38: Why is this definition necessary since there already exists a `VerificationMethod` definition?

src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts
```diff
@@ -0,0 +1,205 @@
...
+  id: string
+  alsoKnownAs?: string[]
+  controller?: string | string[]
+  verificationMethod?: VerificationMethodLegacyCommonSigner[]
+  service?: ServiceEndpoint[]
+  publicKey?: VerificationMethodLegacyCommonSigner[]
+  authentication?: VerificationMethodLegacyCommonSigner[]
+}
+
+type DIDDocumentCommonSigner = {
```
mirceanis, line 62: Why is this necessary?

src/SignerAlgorithm.ts
bshambaugh: Understandable. It might have a bit to do with how my mind was working. I grasped things as, and after, I worked through them. I realize it was a huge refactor. It seemed necessary as otherwise the files were getting to get huge combining multiple signers and it would be hard to follow. I told Joel that I was trying to figure things out for my particular project. I need a remote signer over a radio connection. I see mention of constructing a remote signer. I went ahead and added ES256 and walked through the library so I'd have it if I need it when bringing in https://github.com/decentralized-identity/did-jwt-vc later. I'm also trying to build something like this (https://github.com/ceramicnetwork/key-did-provider-ed25519) but for a remote signer. Yeah, I was kind of aggressive with my commit. I think I have a got to get this done mentality. I've been working on a particular (presently unfunded/hobby/or whatever project) for over a year, and get onto myself for not making enough progress to be successful with it. I'll be patient. Maybe in time I'll discover that I did some work that did not need to be done. I realize other people have to deal with what I put out. Maybe I'm at a point where I can talk intelligently about it, if I did not realize before...

mirceanis: no need to apologise :)

The bill of changes will be reduced once you remove `package-lock.json` since this project uses `yarn` instead, and further more after removing `ES256-R` and related code.

It might not be worth trying to adapt the blockchain/cosmos bits since AFAIK there is no use of P-256 on the blockchains involved, only secp256k.

Still, there is very much to review, and it will take very long.

kimdhamilton:
I am philosophically tremendously supportive of this week, but also need time to review. For reference, I believe this will also help us with satisfying more of the did-jwt-vc tests (https://github.com/decentralized-identity/JWS-Test-Suite).

OR13:
This library will need support for P-256 / ES256 or P-384 / ES384... especially if it needs to support at least 1 NIST approved signature format.

package.json
```diff
@@ -89,7 +91,7 @@
     "elliptic": "^6.5.4",
     "js-sha3": "^0.8.0",
     "multiformats": "^9.4.10",
-    "uint8arrays": "^3.0.0"
+    "uint8arrays": "^3.0.0",
```
e-y, line 94: Remove comma here. Causes yarn and npm JSON parsing error.

bshambaugh: Thanks for noticing this. I saw this too. package.json now parses.

bshambaugh:
> This library will need support for P-256 / ES256 or P-384 / ES384... especially if it needs to support at least 1 NIST approved signature format.

I added an issue that provides a sampling of where things would be different: https://github.com/decentralized-identity/did-jwt/issues/219 . It is my belief that this re-organization in this PR could support ES384, ES512, RSA, etc...

bshambaugh:
@decentralgabe is also working with the JWS-Test-Suite IIRC. He may find this PR interesting?  Well anyway, I may have to come back to this in 10 days or so.

src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts
bshambaugh: Short answer. I found the did-resolver library limited in terms of its interfaces and wanted to package the creation of did documents in functions. It seemed like a bit of a hack. I will get back to this. Thank you.

mirceanis (changes requested):
please remove package-lock.json as this project uses yarn.lock.
This will also to make this review less scary.

src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts
mirceanis: can you be more specific in where the interfaces are limiting?

The interfaces defined there are meant to mimic the DID spec, with some support for previous versions of the spec. If they are limiting, this might be to match the spec, but, if we missed something there, please raise an issue, instead of redefining these types.

bshambaugh: There were some other things in there like (https://github.com/bshambaugh/did-jwt/blob/master/src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts#L24-L75) that were added to avoid the use of any in the typescript file. This might need to be refactored. I guess I didn't realize how scary this was going to get.

src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts
bshambaugh: > There were some other things in there like (https://github.com/bshambaugh/did-jwt/blob/master/src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts#L24-L75) that were added to avoid the use of **any** in the typescript file. This might need to be refactored. I guess I didn't realize how scary this was going to get.

export interface VerificationMethod {
  id: string
  type: string
  controller: string
  publicKeyBase58?: string
  publicKeyBase64?: string
  publicKeyJwk?: JsonWebKey
  publicKeyHex?: string
  publicKeyMultibase?: string
  blockchainAccountId?: string
  ethereumAddress?: string
}

https://github.com/decentralized-identity/did-resolver/blob/master/src/resolver.ts#L88-L99

I recall that the legacy versions were not matching the https://www.w3.org/TR/did-core/ spec.

mirceanis: The reason I'm asking this is because the interface is already exported by did-resolver and you can extend it for the legacy use-case tests, making it more obvious which bits are considered legacy.

This is more a matter of preference, honestly, since we're dealing with tests here.

src/VerifierAlg/ES256KVerifierAlg.ts
```diff
@@ -0,0 +1,91 @@
...
+import { sha256, toEthereumAddress } from '../Digest'
+import type { VerificationMethod } from 'did-resolver'
+import { bytesToHex, EcdsaSignature } from '../util'
+import { verifyBlockchainAccountId } from '../blockchains'
+import * as CommonVerifierAlg from './CommonVerifierAlg'
+
+const secp256k1 = new EC('secp256k1')
+
+/*
+interface LegacyVerificationMethod extends VerificationMethod {
```
mirceanis, line 11: I prefer this simple extension of the publicly defined type as opposed to redefining the interfaces all together.
It seems clearer about what the legacy bits are.

src/VerifierAlgorithm.ts
mirceanis: just to be clearer, all these imports should be simplified this way, not just this line

src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts
mirceanis: The use of `any` and `unknown` in the test files is very much acceptable, as well as type casting for the sake of making tests simple to understand, especially for testing corner cases, as long as there exist some tests that run through the types as they are expressed or expected by the API.

You shouldn't feel the need to redefine big chunks of API just for testing corner cases.

Since these libraries are ultimately boiled down to plain JavaScript, it is even better to have tests that don't fit the neatly defined types that the IDE can warn you about.

src/__tests__/JWT/CommonSignerTest/__tests__/CommonSignerTest.test.ts
```diff
@@ -0,0 +1,78 @@
...
+     const pemPublic = jwkToPem.default(publicToJWK(publicPointHex_x,publicPointHex_y,'EC','P-256'));
+   
+     const isvalid  = verifyTokenFormAndValidity(token,pemPublic)
+     
+     expect(verifyTokenFormAndValidity(token,pemPublic)).toBe(true); 
+ })
+
+describe('SignerAlgorithm', () => {
+
+  it("returns a did doc legacy", () => {
```
mirceanis, line 66: I appreciate the thoroughness, but it feels like overkill to test the methods used during tests :)

src/__tests__/SignerAlg/ES256SignerAlg.test.ts
```diff
@@ -0,0 +1,87 @@
...
+const kp = secp256r1.keyFromPrivate(privateKey)
+const signer = ES256Signer(privateKey)
+const recoverySigner = ES256Signer(privateKey, true)
+
+describe('SignerAlgorithm', () => {
+  it('supports ES256', () => {
+    expect(typeof SignerAlgorithm('ES256')).toEqual('function')
+  })
+
+  it('supports ES256-R', () => {
```
mirceanis, line 18: please remove `ES256-R`

mirceanis (changes requested):
I made a bit more progress on the review.

Still, this is going extremely slowly due to the massive refactoring of the tests. It would be much easier to review if we would have the original tests as intact as possible, and update them in a separate PR.
I am very tempted to close this PR and request that the tests are reverted, this is a huge time drain :(

src/JWT.ts
bshambaugh: Okay, I have the code for ES256-R. I can archive it for now, and reconsider it later. (?) The reason for this PR was to support something like https://github.com/bshambaugh/key-did-provider-p256/blob/master/src/index.ts . This is a non-functional sketch (at the moment) to support EIP-2844 (https://eips.ethereum.org/EIPS/eip-2844) . At the EthDenver Hackathon I heard about https://f-o-a-m.github.io/foam.developer/foamlite/end-node.html which mentions use of a recovery method that grabs the public key based on a signature. (I discovered that they are also using LoRa.) I literally just read this 5 or 10 minutes ago so I am not educated in the use nor in foam. I think that their protocol is even lower level than what I am attempting.

https://www.youtube.com/watch?v=Z8Wf7Srsg5U&t=50m20s is a Ceramic Community Call that mentions Joel's (oed's) introduction of EIP-2844 to my problem which sent me down this road. Since this (described in the Call) is a preexisting project I may not apply it (or may not be allowed to) to the EthDenver hackathon. In the next week or so, I may in fact try something else. For context, this is the project I was mentioning in the Community Call: https://github.com/bshambaugh/BlinkyProject .

bshambaugh: > I agree with @oed, `ES256-R` is not needed. `ES256K-R` existed as a pattern and was in use before it got a name. It exists now to allow for `blockchainAccountId` verification methods to be used as verifiers. We could probably phase it out of existence if we accept 1 bit less of security for ES256K + blockchainAccountId

I'll have to peruse the code to understand the context. Since Joel knows about what I have been up to, I assume I can just prune it out for now.

src/util.ts
```diff
@@ -85,8 +85,8 @@ export function toSealed(ciphertext: string, tag: string): Uint8Array {
 }
 
 const hexMatcher = /^(0x)?([a-fA-F0-9]{64}|[a-fA-F0-9]{128})$/
-const base58Matcher = /^([1-9A-HJ-NP-Za-km-z]{44}|[1-9A-HJ-NP-Za-km-z]{88})$/
-const base64Matcher = /^([0-9a-zA-Z=\-_+/]{43}|[0-9a-zA-Z=\-_+/]{86})(={0,2})$/
+const base58Matcher = /^([1-9A-HJ-NP-Za-km-z]{43,44}|[1-9A-HJ-NP-Za-km-z]{86,88})$/
```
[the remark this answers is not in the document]
bshambaugh: Sometimes the base58 and bas64 matchers were conflicting as I recall. Here is the transcript:
/ debug log:
https://gist.github.com/bshambaugh/d0f87d471040399818eb7d258b660c49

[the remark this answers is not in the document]
mirceanis: ok, if this is a problem let's address it in a separate issue describing the bug and with a test vector that fails with the existing matchers. Please revert this for this PR.

The keys being matched are of fixed size, there is no reason to have varying length here.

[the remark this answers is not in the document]
mirceanis: The issue here is the change to the base58 matcher.
Please describe why this change was necessary with an example that fails with the current matcher.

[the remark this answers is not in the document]
bshambaugh: I created an issue: https://github.com/decentralized-identity/did-jwt/issues/222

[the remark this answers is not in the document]
bshambaugh: I'll try to clean this up soon. My coffee ran out and it let me to let up on it a lot. As a result, I have really bad brain fog at the moment,and cannot really think clearly.

src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts
bshambaugh: I'm in the process of going through this, and thus made some changes to CommonSignerTest,ts, and could possibly do more. I posted a description of what I was up to in a slack channel, which helped guide my mind:
"CommonSignerTest.ts is an interesting file. It is a refactoring of part of https://github.com/decentralized-identity/did-jwt/blob/master/src/__tests__/JWT.test.ts . I split it out, while refactoring (the test, grin...).
Line 36 through 91 are interesting, and they were reformulated for use in a common file. Here is the new structure: https://github.com/bshambaugh/did-jwt/tree/master/src/__tests__/JWT . The const declarations that were
formally didDocLegacy, didDoc, and audDidDoc are now functions in ./CommonSignerTest/CommonSignerTest.ts in the new structure and CommonSignerTest.ts is called by JWT.ES256KSigner.test.ts and JWT.ES256Signer.test.ts .
I did this to avoid bloating of the test orignal test file. This makes my reviewer a bit bewildered. He is no longer sure what to trust. (https://github.com/decentralized-identity/did-jwt/pull/212#pullrequestreview-903264857).Along the way the typescript editor in the Visual Studio IDE started complaining about my use of any. Interfaces came to the rescue. I was not entirely satisfied with the interfaces that came with the did-resolver npm package.
(https://github.com/decentralized-identity/did-resolver/blob/master/src/resolver.ts) , so I created a few of my own (https://github.com/bshambaugh/did-jwt/blob/master/src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts#L20-L36)
for legacy verification method and the private key (reason given here: https://github.com/decentralized-identity/did-resolver/issues/121).
I also had to create some types for the legacy and more current did documents (https://github.com/bshambaugh/did-jwt/blob/master/src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts#L38-L62). In addition, I had to re-declare the
JsonWebKey interface since it was not exported from the did-resolver npm package (https://github.com/bshambaugh/did-jwt/blob/master/src/__tests__/JWT/CommonSignerTest/CommonSignerTest.ts#L6-L18)."

bshambaugh: tl;dr: maybe I should focus in on this:

"as well as type casting for the sake of making tests simple to understand, especially for testing corner cases, as long as there exist some tests that run through the types as they are expressed or expected by the API."

Do I need to write things in a different way? I had type any for CreatedidDocLegacy , CreatedidDoc and CreateauddidDoc , but went the interface route when I got complaints from my IDE.

bshambaugh: Okay, CommonSignerTest.ts  [1] is substantially simplified  [1] (https://github.com/decentralized-identity/did-jwt/pull/212/files#diff-0a517e08539420acec2253799dfa947d00c7d2e854614a743dd72a08b9fd5664)

bshambaugh: @mirceanis , CommonSignerTest.ts is substantially cleaned up. I reused declarations from did-resolver where possible.

bshambaugh:
Okay, fair enough. I'll close this pull request and archive the code I have done. I have also been considering JSON Web Encryption that is not included in this PR. See RFC7516 (https://datatracker.ietf.org/doc/html/rfc7516) and https://github.com/decentralized-identity/did-jwt/blob/master/src/xc20pEncryption.ts .

mirceanis:
Thank you!
I look forward to your next PR.

Regarding JWE, there is plenty to be improved in that area as well. Please raise issues before posting PRs, though.

In terms of refactoring, you have good ideas for splitting code and we will need that for #170

Thank you for your work!

Issue #216, opened 2022-01-27
webpack 5: ReferenceError: Buffer is not defined


alerios:
### Current Behavior
When using did-jwt with webpack 5, the following error shows up: "ReferenceError: Buffer is not defined", since Buffer is not included in the browser.

### Failure Information

There are several references to the use of "Buffer.from" in the did-jwt code: https://github.com/decentralized-identity/did-jwt/search?q=%22Buffer.from%22

#### Alternatives you considered

Based on what I could find, these could be replaced to use "u8a.fromString" instead for better compatibility.

In the meantime I applied a polyfill using this guide, which cleared the issue: https://viglucci.io/how-to-polyfill-buffer-with-webpack-5

---

> **Don't paste private keys anywhere public!**

mirceanis:
Thanks for reporting this!
Since you already suggested a fix and a workaround, do you want to create a PR with the fix?

ukstv:
There is now a PR to fix this: https://github.com/decentralized-identity/did-jwt/pull/217

PR #217, opened 2022-01-27
feat: Drop Buffer

ukstv:
Should fix https://github.com/decentralized-identity/did-jwt/issues/216


mirceanis (approved):
looks great, thanks for the fast fix!

PR #240, opened 2022-08-05
Add ES256Signer

bshambaugh:
This is a redo of PR #212 . For the time being, I am changing 1 file and adding 2. There will be more. I figure just changing 3 at first will be more digestible.


mirceanis (changes requested):
This is wonderful and kudos for the approach with smaller PRs that are easier to digest :)

There is a tiny linter error.
Please run `yarn run format --fix` and commit the tiny changes. Everything else looks great!

Issue #239, opened 2022-08-01
[proposal] Improvements to the verification API


mirceanis:
**Is your feature request related to a problem? Please describe.**

The verification of JWT/JWS uses exceptions to signal failures, but this creates a problem in distinguishing between legitimate failures (bad signature, invalid audience, timestamp issue, etc) and exceptional situations (network timeout while resolving DID, malformed input, unsupported algorithm, etc).

I propose separating these 2 scenarios internally so that users get a better idea about what failed and so that control flow for invalid JWTs no longer relies on parsing error messages.

**Describe the solution you'd like**
The result of `verifyJWT()` should be an object with a `verified` property set to true only when signature, timestamps, and audience are valid, and false otherwise.
When one of the validation criteria fails, the reason can be returned as an error code with a message.
`verifyJWT()` should not throw errors.

The verification options should accept a `now` parameter for checks involving past/future validity, or allow the user to disable timestamp checking. Similarly, it should be possible to skip audience checking (although, this is already indirectly possible by pre-inspecting the desired audience).

**Describe alternatives you've considered**
The desired behavior can already be achieved by calling `verifyJWS` directly for signature checks and by performing timestamp and audience checking out of band, but that defeats the purpose of having a library to do this.

Another acceptable solution would be to introduce a new method (`verifyJWTSafe`) that can give detailed reasons for failure without using exceptions for flow control, while leaving the existing `verifyJWT` method intact for backward compatibility.

Please add your opinion about this subject if you have one.

PR #241, opened 2022-08-10
Introduced Policies to override Timestamp checking

daniyal-khalil:
### Description
This merge requests introduces a new interface in the JWT.ts module which JWTVerifyPolicies. The policies at the moment are only restricted to timestamps, but can be further expanded upon to serve better. The JWTVerifyPolicies are made a part of VerifyJWTOptions defined as `policies`. These policies are used to be able to override a certain checks that are made in the VerifyJWT command. It is not a breaking change but an extension to the current implementation.

### Policies
The checks for these policies will not me made only and only if they are explicitly stated to be 
- nbf  (Not Before)
- iat   (Issued At)
- exp (Expires At)

### Tests
Further tests were added to the JWT.test.ts to check whether the implemented policies were properly allowing the presentation to verify in case of a failure and the policy was made to be overridden. The following cases are covered:
- nbf is in the future and policy for nbf is false
- nbf is in the future and iat is in the past with nbf policy false
- expired JWT with exp policy false
- expired JWT without skew time but exp policy false
- iat is in the future with iat policy false

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -38,9 +38,9 @@ export interface JWTVerifyOptions {
 }
 
 export interface JWTVerifyPolicies {
-  now?: boolean,
-  nbf?: boolean,
-  iat?: boolean,
+  now?: boolean
+  nbf?: boolean
+  iat?: boolean
   exp?: boolean
 }
 
@@ -338,8 +338,8 @@ export async function verifyJWT(
       nbf: undefined,
       iat: undefined,
       exp: undefined,
-      now: undefined
-    }
+      now: undefined,
+    },
   }
 ): Promise<JWTVerified> {
   if (!options.resolver) throw new Error('missing_resolver: No DID resolver has been configured')
@@ -349,7 +349,7 @@ export async function verifyJWT(
       ? 'authentication'
       : undefined
     : options.proofPurpose
-  
+
   let did = ''
 
   if (!payload.iss) {
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -34,6 +34,13 @@ export interface JWTVerifyOptions {
   skewTime?: number
   /** See https://www.w3.org/TR/did-spec-registries/#verification-relationships */
   proofPurpose?: ProofPurposeTypes
+  policies?: JWTVerifyPolicies
+}
+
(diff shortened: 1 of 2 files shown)


mirceanis (changes requested):
This looks good and is a good first pass at #239.
Please take a moment to comment on that too if you have opinions about it. IMO, using a `now` parameter for the policy to represent the timestamp at which the JWT should be valid is also a good way to override some timestamp checks, with a more precise intention, and these 2 methods don't exclude each other.. one can specify a `now` and also disable a `nbf` check in the same policy.

Regarding this PR, there are some linter errors which should be easy to fix. Please run `yarn lint --fix` and commit the changes.

daniyal-khalil:
@mirceanis 
Thank you for such a fast response, I have added the `now` policy which looks at all the timestamp checks, if it is set to false, the verification will not look at the timestamps at all. Similarly, as mentioned above, `now` can be true, while `nbf` can still be false. Let me know if I misunderstood anything regarding this, Thank you so much!

src/JWT.ts
```diff
@@ -363,15 +389,17 @@ export async function verifyJWT(
   const skewTime = typeof options.skewTime !== 'undefined' && options.skewTime >= 0 ? options.skewTime : NBF_SKEW
   if (signer) {
     const nowSkewed = now + skewTime
```
mirceanis, line 391: I think you misunderstood what I meant by the `now` parameter.
I was referring to an override to the internal now, which is currently computed as `const now: number = Math.floor(Date.now() / 1000)`

My proposal was that the policy can contain a `now` override as a `number`, not a `boolean`.
This number would be used instead of `new Date().now()/1000` and should represent a timestamp.

mirceanis (changes requested):
Kudos for the thorough testing work!

Please read my comment about the now param. Yeah, there was a bit of a misunderstanding.

src/JWT.ts
daniyal-khalil: Sorry for misunderstanding, fix it and replaced the test case for it as well. Thank you!

mirceanis (approved):
Looks great, thank you for contributing this!

PR #242, opened 2022-08-15
feat: add `aud` override policy for verification

### LINKED ISSUE #239: [proposal] Improvements to the verification API

mirceanis:
fixes #239 
.. without the mentioned breaking change.
Errors are still thrown, but the prefixes for the known failure cases are also exported now as an enum.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 [![npm](https://img.shields.io/npm/dt/did-jwt.svg)](https://www.npmjs.com/package/did-jwt)
 [![npm](https://img.shields.io/npm/v/did-jwt.svg)](https://www.npmjs.com/package/did-jwt)
-[![Twitter Follow](https://img.shields.io/twitter/follow/uport_me.svg?style=social&label=Follow)](https://twitter.com/uport_me)
+[![Twitter Follow](https://img.shields.io/twitter/follow/veramolabs.svg?style=social&label=Follow)](https://twitter.com/veramolabs)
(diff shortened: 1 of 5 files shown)


simonas-notcat (approved):

nickreynolds:
beautiful!

PR #246, opened 2022-09-02
Recoverable parameter for ES256 may be invalid due to EIP-2098. Hence it was removed.

bshambaugh:
There was an issue with https://github.com/decentralized-identity/did-jwt/pull/240 . The recovery parameter may be invalid because the library relies on key recovery from an Ethereum Address. The only valid curve type for an Ethereum address may be secp256k1. See EIP-2098: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2098.md .

The recoverable parameter was removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [6.6.0](https://github.com/decentralized-identity/did-jwt/compare/6.5.0...6.6.0) (2022-08-19)
+
+
+### Features
+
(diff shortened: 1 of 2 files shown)


mirceanis (changes requested):
Looks good, indeed ES256 signer should not produce a recovery param. This is probably unrelated to EIP2098, though.

Can you please resolve the conflicts so we can merge this in?

PR #262, opened 2022-12-08
add support for SIOP request JWT

siacomuzzi:
According to JWT Presentation Profile, the SIOP Request JWT does not contain an `iss` claim, it must have a `client_id` claim instead. See https://identity.foundation/jwt-vc-presentation-profile/#self-issued-op-request-object

image

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -402,7 +402,7 @@ export async function verifyJWT(
       : undefined
     : options.proofPurpose
 
-  let did = ''
+  let did
 
   if (!payload.iss && !payload.client_id) {
     throw new Error(`${JWT_ERROR.INVALID_JWT}: JWT iss or client_id are required`)
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -422,8 +422,15 @@ export async function verifyJWT(
       throw new Error(`${JWT_ERROR.INVALID_JWT}: JWT did is required`)
     }
     did = payload.did
+  } else if (payload.scope === 'openid' && payload.redirect_uri) {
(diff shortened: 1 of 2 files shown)


src/JWT.ts
```diff
@@ -423,7 +423,7 @@ export async function verifyJWT(
     }
     did = payload.did
   } else {
-    did = payload.iss
+    did = payload.iss || payload.client_id
```
OR13, line 426: What if `iss` and `client_id` are both present and non equivalent.

OR13 (changes requested):
prefer to see handling of `iss` and `client_id` alignment or misalignment.

src/JWT.ts
siacomuzzi: added an IF statement just for this kind of JWT

siacomuzzi: > What if `iss` and `client_id` are both present and not equivalent.

According to the spec, `iss` shouldn't be present. So, if both `iss` and `client_id` are present, this proposed logic shouldn't be reached.
The tricky part here is that there is no a direct way to detect if payload belongs to a JWT Presentation Profile request.

mirceanis (approved):
Looks great!

Thank you for adding this.

siacomuzzi:
> Looks great!
> 
> Thank you for adding this.

Thank you for your review! Do you have plans to publish a new release that includes this change soon?

mirceanis:
> > Looks great!
> > Thank you for adding this.
> 
> Thank you for your review! Do you have plans to publish a new release that includes this change soon?

Of course

PR #269, opened 2023-02-13
build: Added ESM types exports resolution

Eengineer1:
- Adds missing type exports for `nodenext` resolution.


mirceanis (approved):
Looks good, thanks for adding this!

PR #271, opened 2023-03-06
fix: provide hash algorithms as exported functions

klaus-sap:
Call implementations of hash functions from native crypto library


mirceanis:
This library is used in contexts other than nodejs.
Will this change also work in browser and react-native environments?

klaus-sap:
Good point, thank you. I will update this PR

mirceanis (approved):
looks ok to me

PR #272, opened 2023-03-07
Added conditional logic following the ConditionalProof specification

theblockstalk:
Adds the ability to verify JWTs with multi-signatures and delegated signatures and combinations of the two, against DIDs that follow the ConditionalProof specification.

https://github.com/w3c-ccg/verifiable-conditions

See slide deck for overview of changes: https://docs.google.com/presentation/d/13jaeireCRFoI1jFPFkbwBhTMQgUG7zEHRjmq0PWz-CI


src/JWT.ts
```diff
@@ -326,16 +343,94 @@ export async function createJWT(
...
+
+    payloadResult = { jwt }
+  }
+  return jwt
 }
 
-function verifyJWSDecoded(
-  { header, data, signature }: JWSDecoded,
+export function verifyJWSDecoded(
+  { header, payload, data, signature }: JWTDecoded,
```
mirceanis, line 404: This should be reverted as it would change the expected input for `verifyJWS`, which is an exported method.

`verifyJWS*` methods shouldn't expect the payload to decode to anything, and just treat it as a blob of bytes.
This would also remove the need to cast the data types in weird ways (`jwsDecoded as unknown as JWTDecoded`)
The logic you added should be performed at a higher level. Perhaps in `verifyJWT`?

src/VerifierAlgorithm.ts
```diff
@@ -78,7 +78,7 @@ export function verifyES256(data: string, signature: string, authenticators: Ver
 
   const signer: VerificationMethod | undefined = fullPublicKeys.find((pk: VerificationMethod) => {
     try {
-      const pubBytes = extractPublicKeyBytes(pk)
+      const pubBytes = extractPublicKeyBytes(pk) as Uint8Array
```
mirceanis, line 81: I wonder why this explicit type was needed here. Can you clarify?

src/JWT.ts
```diff
@@ -351,7 +446,7 @@ function verifyJWSDecoded(
  */
 export function verifyJWS(jws: string, pubKeys: VerificationMethod | VerificationMethod[]): VerificationMethod {
   const jwsDecoded: JWSDecoded = decodeJWS(jws)
-  return verifyJWSDecoded(jwsDecoded, pubKeys)
+  return verifyJWSDecoded(jwsDecoded as unknown as JWTDecoded, pubKeys)
```
mirceanis, line 449: This seems unsafe.
`verifyJWS` is an exported function. Its users don't expect the payload to be decodable.

mirceanis (changes requested):
Apologies for coming back to this review so late.

It seems that `verifyJWS` functionality is being changed with the expectation that the data is a JWT payload.
This is not a valid assumption and since it is an exported function the changes related to that should be reverted.

src/VerifierAlgorithm.ts
theblockstalk: This was not needed. I have removed it.

src/JWT.ts
theblockstalk: I have now reverted `verifyJWSDecoded()` back to original function and created a new function `verifyJWTDecoded()` instead with the required changes.

src/JWT.ts
theblockstalk: I have removed this typing

theblockstalk:
I have made changes per your comments.

I have also added test suite `ConditionalAlgorithm.test.ts` which covers many scenarios. Test suite is now passing all tests.

src/JWT.ts
```diff
@@ -392,61 +493,99 @@ export async function verifyJWT(
...
+    ))
+    // Add to options object for recursive reference
+    options.didAuthenticator = { didResolutionResult, authenticators, issuer }
+  }
+
+  const { did } = parse(didUrl) as ParsedDID
+
+  let signer: VerificationMethod | null = null
+
+  if (did !== didUrl) {
```
mirceanis, line 564: This comparison assumes that a didUrl that does not match the parsed DID refers to one of the verification methods.
The assumption will break if the didUrl contains, for example, query parameters.

Example: `did:ethr:0xabcd...?versionId=1234`.
This is a valid issuer with a DID document potentially different from `did:ethr:0xabcd...`

Is the intent here to avoid looping through the `authenticators` array?

mirceanis (commented):
This looks good.

I think there's a potential error introduced from trying to parse the issuer DID.
Can you clarify the intent there?

src/JWT.ts
theblockstalk: if the issuer of the DID is a DID URL, here we assume that it is pointing to a specific verification method in the DID Document. This is used in the recursive calling to `verifyJWT()` for delegation

`did:ethr` is not a good example, as it cannot express multi-sig conditions or delegations.

Take for example two DIDs which one verification method on each:
1. `did:antelope:jack#delegated` which delegates to `did:antelope:thomas#base`
2. `did:antelope:thomas#base` with a ES256K keypair

A VC is issued by `did:antelope:jack#delegated`. The VC is signed by the  `did:antelope:thomas#base` keypair.

- `verifyJWT()` is called which finds the issuer to be `did:antelope:jack#delegated`
- it goes and finds the specific verification method, which tells that it is delegating to `did:antelope:thomas#base`
- `verifyJWT()` is then called recursively to check whether the signature matches the `did:antelope:thomas#base` which it does.

It is important to note that at each of the above steps, a specific verification method is found using the DID URL, required to make the above flow work.

mirceanis: Thank you for clarifying the intent.
I realize did:ethr is not relevant to the conditional verification, but the changes I'm asking about are relevant for all verification, not just conditional.

I'll add a test for the case I'm referring to.

theblockstalk: Yes I see your consern. If someone issues using a DIDUrl that does not point directly to a verification method, it will now fail. Correct?

theblockstalk: I'm thinking of ways around this, perhaps we can change it so that it only checks the fragment part of the DID URL against the fragment of the verification method. What do you think?

mirceanis: That's one possibility.
Another option would be to just try to find the verification method, which would work for the conditional logic you need, but then to not fail if it's not found, which would continue to do the verification as before.

theblockstalk: JWT.ts verifyJWT()

```ts
  const { fragment } = parse(didUrl) as ParsedDID

  let signer: VerificationMethod | null = null

  // If a fragment is supplied, assume it is pointing to a specific Verification Method
  // This is a required feature for Conditional Proof delegated signatures
  if (fragment) {
    const authenticator = authenticators.find((auth) => (parse(auth.id) as ParsedDID).fragment === fragment)
 ```

this would avoid this situation (not 100%, e.g. if someone used fragments to refer to a non verification method it would still break)

Comment (mirceanis) [reply src/JWT.ts:564]: Your solution with the fragment seems like the right approach, but somehow comparing just the fragment parts seems odd, as it seems close to being able to impersonate some other DID.

How about this pseudocode?
```
if (didURL contains fragment) {
  find authenticator that matches didUrl
} else {
  try to check the proof against all authenticators
}
```

theblockstalk: the original intent, was to treat the DID URL as a relative URL, which are assumed to dereference to a resource in the DID Document

mirceanis: OK, that's even better, as it would also cover the case where the issuer is a URL but there's also a `kid` specified in the header, relative to the issuer URL.

I'll merge this PR as is and make the relative `kid` changes as a separate item.

mirceanis (approved):

PR #274, opened 2023-03-15
chore: remove dev dependency did-key-creator

nickreynolds:
- this dev dependency was bringing in versions of multiformats / multicodec that may be causing issues in other projects
- it's easiest to just remove this dependency and hardcode the values it was generating, rather than sorting out all dependency issues across multiple repos

### DIFF
--- a/src/__tests__/JWT.test.ts
+++ b/src/__tests__/JWT.test.ts
@@ -155,10 +155,10 @@ describe('createJWT()', () => {
 
     // input public key in hex, and export pem
     function publicToJWK(
-    publicPointHex_x: string,
-    publicPointHex_y: string,
-    kty_value: string,
-    crv_value: string
+      publicPointHex_x: string,
+      publicPointHex_y: string,
+      kty_value: string,
+      crv_value: string
     ): JsonWebKey {
     if (publicPointHex_x.length % 2 != 0) {
       publicPointHex_x = '0' + publicPointHex_x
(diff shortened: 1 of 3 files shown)


src/__tests__/JWT.test.ts
```diff
@@ -121,151 +120,156 @@ const audDidDoc = {
 
 describe('createJWT()', () => {
   describe('ES256', () => {
-        const alg = 'ES256'
```
nickreynolds, line 124: this is just whitespace changes

src/__tests__/JWT.test.ts
```diff
@@ -121,151 +120,156 @@ const audDidDoc = {
...
-          crv: crv_value,
-          d: privatePointBase64URL,
-        }
+    // construct did:key for secp256r1 (unlike did for secp256k1 which is for an Ethereum Address)
+    // This originally was constructed by `encodeDIDfromHextString` imported from `did-key-creator`
+    // package, but that dependency was removed, so now `did` is just hardcoded
+    // const multicodecName = 'p256-pub';
+    // const publicKey = '0314c58e581c7656ba153195669fe4ce53ff78dd5ede60a4039771a90c58cb41de'
+    // const did = encodeDIDfromHexString(multicodecName,publicKey)
+    const did = 'did:key:zDnaej4NHntda4rNW4FBUJgFzdcgEAXKGRVGE8LuVfRbuMuc1'
```
nickreynolds, line 135: this is the actual change

mirceanis (approved):
Looks great, thanks for clearing up some dependencies

PR #284, opened 2023-05-17
feat: refactor JWE support to allow external algorithm implementations

mirceanis:
This PR introduces the `createFullEncrypter` method that should allow users to provide their own implementations for JWE content encryption or key wrapping.
A few data types were defined to support this, representing objects that provide this functionality along with a description of the algorithm they implement.:  `KekCreator` to create key-wrapping keys, `KeyWrapper` to wrap the content encryption key, `ContentEncrypter` to actually encrypt the payload.

This also deprecates some of the method names used to create encrypters/decrypters in favor of the more complete names that include the actual crypto involved.

This repo went through a lot of refactoring:
* using ESM style relative imports ( adding '.js' at the end)
* bump most dependencies
* prepare jest for pure ESM
* use methods from utils instead of always importing uint8arrays

Example:
```ts
// AuthEncrypter using x25519 ECDH-1PU (v3) key agreement With A256KW key wrapping and XC20P content encryption
const encrypter = createFullEncrypter(
    recipientPublicKey,
    senderSecret,
    { kid: 'did:example:asdf#1234', apu: 'producer data', apv: 'consumer data' }, // options
    { alg: 'ECDH-1PU', createKek: createX25519Ecdh1PUv3Kek }, // kek creator
    { alg: 'A256KW', from: (wrappingKey: Uint8Array) => ({ wrap: async (cek: Uint8Array) => ({ ciphertext: new AESKW(wrappingKey).wrapKey(cek) })   })}, // key wrapper (A256KW)
    { enc: 'XC20P', from: (cek: Uint8Array) => xc20pDirEncrypter(cek) } // content encryptor
  )

const jwe = await createJWE(messageBytes, [encrypter], protectedHeader, undefined, true)
```

This PR would also close #265 #266 #251 and re-enable the automatic dependency management that was blocked by them.


simonas-notcat (approved):

nickreynolds (approved):
Looks great! Do you think in a future (smaller) refactor, it would make sense to change some of the strings (e.g. "X25519KeyAgreementKey2019") to be constants in some shared file?

mirceanis:
> Looks great! Do you think in a future (smaller) refactor, it would make sense to change some of the strings (e.g. "X25519KeyAgreementKey2019") to be constants in some shared file?

Of course. There's actually a lot more to be refactor here to make the code more friendly, modular and safer. It's just a question of priorities

Issue #225, opened 2022-04-07
JWE for P-256


bshambaugh:
I see https://w3c-ccg.github.io/lds-jws2020/ .

This suite support cryptographic agility, see [[RFC7696](https://w3c-ccg.github.io/lds-jws2020/#bib-rfc7696)]. This table maps a key type to a subset of [[IANA_JOSE](https://w3c-ccg.github.io/lds-jws2020/#bib-iana_jose)] supported signing and encryption algorithms.
kty 	crvOrSize 	signature 	keyAgreement 	encryption
OKP 	Ed25519 	EdDSA 		ECDH-ES+A256KW
OKP 	X25519 		ECDH 	ECDH-ES+A256KW
EC 	secp256k1 	ES256K 	ECDH 	ECDH-ES+A256KW
**EC 	P-256 	ES256 	ECDH 	ECDH-ES+A256KW**
EC 	P-384 	ES384 	ECDH 	ECDH-ES+A256KW
RSA 	2048 	PS256 		RSA-OAEP

I would like the stuff in bold because I would like JWE with P-256?

Here is some babbling from earlier:

This is kind of strange looking:

const kek = concatKDF(sharedSecret, keyLen, alg)
const res = xc20pEncrypter(kek)(cek)

function xc20pEncrypter(key: Uint8Array): (cleartext: Uint8Array, aad?: Uint8Array) => EncryptionResult {
  const cipher = new XChaCha20Poly1305(key)
  return (cleartext: Uint8Array, aad?: Uint8Array) => {
    const iv = randomBytes(cipher.nonceLength)
    const sealed = cipher.seal(iv, cleartext, aad)
    return {
      ciphertext: sealed.subarray(0, sealed.length - cipher.tagLength),
      tag: sealed.subarray(sealed.length - cipher.tagLength),
      iv,
    }
  }
}

source: https://github.com/decentralized-identity/did-jwt/blob/master/src/xc20pEncryption.ts#L126-L204

[bshambaugh]
Maybe I should ask with an issue what is needed.I'm trying to reverse engineer the code.https://www.rfc-editor.org/rfc/rfc7518.html#section-4.6I need instead:

   | ECDH-ES+A256KW  | ECDH-ES using Concat KDF and CEK wrapped with   |
   |                 | "A256KW"  

[bshambaugh]
I can get A256KW here: https://github.com/StableLib/stablelib/blob/master/packages/aes-kw/aes-kw.test.ts
[...]

PR #295, opened 2023-09-22
Implementing JWE for P256

bshambaugh:
I am starting to work through #225 and the first file that I changed was ECDH.ts. Please forgive my deletion of yarn.lock. If this is an issue, perhaps I can address it later. This does not have tests yet, but neither does the prexisting x25519 function. I do need to make sure that when I get a compressed P256 public key as input to getSharedSecret I get the correct results.


mirceanis (changes requested):
Looks ok, thank you for adding this.
You need to add the lockfile back before the PR can be accepted.

bshambaugh:
I did git `checkout upstream/master -- yarn.lock` ,  `git add yarn.lock ` , `git commit -m "add removed yarn.lock file"` , and `git push -u origin master ` to get to the present state. I thought that diff for yarn.lock would show now changes, but it still does.

mirceanis:
> I did `git checkout upstream/master -- yarn.lock` , `git add yarn.lock ` , `git commit -m "add removed yarn.lock file"` , and `git push -u origin master ` to get to the present state on my fork. I thought that diff for yarn.lock would show no changes, but it still does.

Perhaps try to `git fetch --all` before the checkout

bshambaugh:
I cloned this repository in a new directory and compared it to the fork and did what you said with git fetch and then did a local diff on yarn.lock. I even browsed through the file on github. The file yarn.lock doesn't appear to be different, yet it is showing differences in files changed for this PR.

bshambaugh:
But, the last change in git log that is not by me is on sept 8th. On this upstream remote, the last change is sept 27th,

bshambaugh:
`git merge origin/master upstream/master` then `git push` did the trick.

mirceanis (commented):
I was trying to work with the p256 shared secret implementation from @noble/curves and discovered that it may have some differences when compared to other implementations (see https://github.com/paulmillr/noble-curves/issues/96)

My instinct here would be to align with the other implementations which would mean skipping the first byte of the output, but ideally this should all be based on some test vectors that ensure interoperability.

BTW, x25519 ECDH doesn't manifest the same problem as the key representation there is less flexible

bshambaugh:
Awesome. I was gone for 2 weeks in the 1st half of October for and IIW et al. trip, but I have started to look at JWE again. This means I have been combing through the code base and searching the web and making notes in the past few weeks. 

tl;dr: good timing. more code may be coming soon.

bshambaugh:
As I was looking through some repositories on didcomm I realized that there was some replication:
https://github.com/decentralized-identity/veramo/tree/next/packages/did-comm/src/encryption

This is not bad because I realized I was doing similar things
Screenshot_2023-11-22_14-04-20

Screenshot_2023-11-22_14-04-34

mirceanis:
> As I was looking through some repositories on didcomm I realized that there was some replication

The code duplication from the did-comm implementation you linked to is not from my proudest moments..

> This is not bad because I realized I was doing similar things

There are definitely improvements to be made, so if you do manage to avoid duplication, that would be even better.


---


also maybe related: https://github.com/paulmillr/noble-ciphers/discussions/18

bshambaugh:
Status update:

I'm trying to refactor code, but I am getting stuck. I apologize for the screenshots, but it seems to allow me to make this concise.

In
https://codesandbox.io/p/github/bshambaugh/did-jwt/master?file=%2Fsrc%2Fencryption%2[…]1%2C1-84%2C1&workspaceId=2a49a562-e345-4cca-bc1f-efecc0ea40b6 

Start also looks like the form in createEncrypterStart.ts

Finish looks like the form in createEncrytperFinal.ts 

start
finsh

createencrypterFinal.ts will be renamed to createEncrypter.ts

bshambaugh:
There are multiple feature branches going on. Number [1]  just keeps the same **alg** and **enc** parameters as fox x25519 but changes the curve to P-256. 
Number [2]  tries to introduce ECDH-ES-A256KW as the **alg** and A256GCM as the **enc**.

For number [1] this still needs to be fixed:
https://github.com/bshambaugh/did-jwt/blob/X25519-ECDH-1PU-ES/src/encryption/xc20pEncryption.ts#L145-L191

For number [2]  a lot still needs to be done. You can see the present @noble and @stablelib having to be there to provide GCM and KW. https://github.com/bshambaugh/did-jwt/blob/as256kwExperiments/src/encryption/a256kwgcm.ts . 
https://github.com/bshambaugh/did-jwt/blob/as256kwExperiments/src/encryption/createEncrypter.ts will need to look like https://github.com/bshambaugh/did-jwt/blob/X25519-ECDH-1PU-ES/src/encryption/createEncrypter.ts ... 

& tests for both.

bshambaugh:
https://github.com/bshambaugh/did-jwt/blob/X25519-ECDH-1PU-ES/src/encryption/createEncrypter.ts#L28 was added

`ephemeralKeyPair?: EphemeralKeyPair`

 to get
https://github.com/bshambaugh/did-jwt/blob/X25519-ECDH-1PU-ES/src/encryption/createEncrypter.ts#L80-L84

` if(ephemeralKeyPair?.publicKeyJWK.crv != null) { //uses added ephemeralKeyPair, the one in encryptCek not in scope
    return { alg: keyWrapper.alg, enc: contentEncrypter.enc, encrypt, encryptCek, genEpk: prefixToDriverMap[ephemeralKeyPair!.publicKeyJWK.crv!] } 
  } else {
    return { alg: keyWrapper.alg, enc: contentEncrypter.enc, encrypt, encryptCek, genEpk: genX25519EphemeralKeyPair }
  }`

to work. Except it doesn't yet. I need a better way to construct a conditional which depends on the cryptographic curve.

I added resolveP256Encrypters which is like resolveX25519Encrypters 

https://github.com/bshambaugh/did-jwt/blob/X25519-ECDH-1PU-ES/src/encryption/xc20pEncryption.ts#L449-L504
https://github.com/bshambaugh/did-jwt/blob/X25519-ECDH-1PU-ES/src/encryption/xc20pEncryption.ts#L145-L191

which is what @oed exported. https://github.com/oed/did-jwt/tree/fix/export-x25519-resolver .

bshambaugh:
I fixed some stuff and All tests on my X25519-ECDH-1PU-ES branch appear to pass now.

src/__tests__/a256gcmEncryption.test.ts
```diff
@@ -925,7 +926,197 @@ describe('ECDH-ES+A256KW (P-256) Auth, Key Wrapping Mode with A256GCM content en
...
+    })
+  })
+})
+
+// Adapted from:
+// https://github.com/decentralized-identity/veramo/blob/next/packages/did-comm/src/__tests__/encryption.test.ts#L492-L674
+// changes X25519 to P256
+describe('ECDH-ES+A256KW (P-256) Dir Auth, Key Wrapping Mode with A256GCM content encryption', () => {
+  describe('One recipient', () => {
+    let cleartext: Uint8Array, recipientKey: any, senderKey: any, decrypter: Decrypter, sharedSecret: Uint8Array
```
bshambaugh, line 945: Comment after DIF Hacakthon: It seems like unauthenticated encryption is the default for JWE with ECDH-ES+A256KW, so I should throw out any code that is '*'.Auth.'*' and just keep ''*'.Anon.'*'? 

i.e.: Keep AnonEncrypter, AnonDecrypter, AnonDirEncrypter and AnonDirDecrypter..

"JSON Object Signing and Encryption (JOSE) defines a number of
   encryption (JWE) [[RFC7516](https://datatracker.ietf.org/doc/html/rfc7516)] and digital signature (JWS) [[RFC7515](https://datatracker.ietf.org/doc/html/rfc7515)]
   algorithms.  When symmetric cryptography is used, JWE provides
   authenticated encryption that ensures both confidentiality and sender
   authentication.  However, for public key cryptography the existing
   JWE encryption algorithms provide only confidentiality and some level
   of ciphertext integrity.  When sender authentication is required,
   users must resort to nested signed-then-encrypted structures, which
   increases the overhead and size of resulting messages.  This document
   describes an alternative encryption algorithm called ECDH-1PU that
   provides public key authenticated encryption, allowing the benefits
   of authenticated encryption to be enjoyed for public key JWE as it
   currently is for symmetric cryptography."
   https://datatracker.ietf.org/doc/html/draft-madden-jose-ecdh-1pu-03

mirceanis: Yes, only `-1PU` is auth, `-ES` is anon

bshambaugh: to-do: make sure https://github.com/decentralized-identity/did-jwt/blob/master/src/__tests__/jwe-vectors.ts is up to date to cover these changes.

bshambaugh:
not stale. still working on.

PR #298, opened 2023-09-27
feat: interpret publicKeyMultibase as multicodec

### LINKED ISSUE #297: [BUG] multicodec encoded publicKeyMultibase is not interpreted correctly
#### Steps to Reproduce

The following test case fails:

```ts
 // resolver that uses publicKeyMultibase ( from the wild )
    const resolver = {
      resolve: async () => ({
        didDocumentMetadata: {},
        didResolutionMetadata: {
          contentType: 'application/did+ld+json',
        },
        didDocument: {
          '@context': ['https://www.w3.org/ns/did/v1', 'https://w3id.org/security/suites/ed25519-2020/v1'],
          id: 'did:peer:0z6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw',
          verificationMethod: [
            {
              id: 'did:peer:0z6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw#6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw',
              type: 'Ed25519VerificationKey2020',
              controller: 'did:peer:0z6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw',
              publicKeyMultibase: 'z6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw',
            },
          ],
          authentication: [
            'did:peer:0z6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw#6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw',
          ],
          assertionMethod: [
            'did:peer:0z6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw#6MknNW5mvrUpssSJwZRQSinLWXzcECPtjzeKUsTR1Mvumfw',
          ],
        },
      }),
    }
    const jwt =
      'eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCJ9.eyJ2YyI6eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvMjAxOC9jcmVkZW50aWFscy92MSJdLCJ0eXBlIjpbIlZlcmlmaWFibGVDcmVkZW50aWFsIl0sImNyZWRlbnRpYWxTdWJqZWN0Ijp7Im5vdGhpbmciOiJlbHNlIG1hdHRlcnMifX0sIm5iZiI6MTY5NTA1MjE4MSwiaXNzIjoiZGlkOnBlZXI6MHo2TWtuTlc1bXZyVXBzc1NKd1pSUVNpbkxXWHpjRUNQdGp6ZUtVc1RSMU12dW1mdyJ9.mvgdqscXYjIXRuut83e8AfcBVdQJJOppQ9flohALoke_qRL9rR0FBOuBjWbf6uHftKv8lqUcqZuPnmsAJ0sbAA'

    // fails with invalid_signature
    const { payload } = await verifyJWT(jwt, { resolver })

```

#### Details

The `extractPublicKeyBytes()` method used to get the raw public key from the VerificationMethod does not account for multicodec.

The publicKeyMultibase in the example is a multibase encoding of a multicodec encoding of the actual key material. This pattern is fairly common, so it should be supported.

mirceanis:
fixes #297

Some users of `publicKeyMultibase` actually encode a multicodec prefixed byte array.
This leads to keys that are not decoded properly and thus are not usable for verification.

This fix includes some heuristics to try to decode multibase directly if it fits known public key lengths and then attempt multicodec decoding.

This PR also exports the methods used for key conversions for use in other libraries.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -2,6 +2,7 @@ import { concat, fromString, toString } from 'uint8arrays'
 import { bases } from 'multiformats/basics'
 import { x25519 } from '@noble/curves/ed25519'
 import type { EphemeralKeyPair } from './encryption/types.js'
+import { varint } from 'multiformats'
 
 const u8a = { toString, fromString, concat }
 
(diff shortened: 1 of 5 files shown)


src/util.ts
```diff
@@ -52,8 +53,81 @@ export function bytesToBase58(b: Uint8Array): string {
...
+ *
+ * @param b - the Uint8Array to be encoded
+ * @param base - the base to use for encoding (defaults to base58btc)
+ * @param codec - the codec to use for encoding (defaults to no codec)
+ *
+ * @returns the multibase encoded string
+ *
+ * @public
+ */
+export function bytesToMultibase(
```
nickreynolds, line 78: would it make sense to eventually pull this function into its own util package? we have a similar function in `@veramo/utils` as you know, and it would be best to not have to maintain 2 identical functions

nickreynolds (approved):

src/util.ts
mirceanis: I'll use the function from here in veramo and remove the duplicate

PR #299, opened 2023-09-27
fix: accept 32 or 64 bytes as keys for EdDSASigner

### LINKED ISSUE #289: [BUG] EdDSA signer accepts a 64 bytes `Uint8Array` while EdDSA private key is 32 Bytes
### Current Behavior

Currently if I go and try to create an EdDSA signer it expects a 64 Byte private key, while EdDSA private key is 32 bytes, I understand it wants to get a concatenated version of Private + Public, however in that case the variable naming seems misleading.

### Expected Behavior

It asks for a 32 Byte key, or the name is more suggestive that you require a concatenated key with priv+public

### Failure Information

Error: bad_key: Invalid private key format. Expecting 64 bytes, but got 32

#### Steps to Reproduce

1. Use a private key seed which is 32 bytes
2. try to create a EdDSA signer
3. observe the signer fail

#### Environment Details

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* node/browser version: v20.2.0
* OS Version: MacOS 13.0.1
* Device: MacBook Pro 2021 M1 Max

#### Alternatives you considered

I would be more than happy to add a pull request which makes it so that we can accept both 32 byte and 64, this way the current functionality doesn't break.

mirceanis:
Only the first 32 bytes are used as the actual key

fixes #289

### DIFF
--- a/src/signers/EdDSASigner.ts
+++ b/src/signers/EdDSASigner.ts
@@ -18,8 +18,8 @@ import { bytesToBase64url, stringToBytes } from '../util.js'
  */
 export function EdDSASigner(secretKey: Uint8Array): Signer {
   const privateKeyBytes: Uint8Array = secretKey
-  if (privateKeyBytes.length !== 64) {
-    throw new Error(`bad_key: Invalid private key format. Expecting 64 bytes, but got ${privateKeyBytes.length}`)
+  if (![32, 64].includes(privateKeyBytes.length)) {
+    throw new Error(`bad_key: Invalid private key format. Expecting 32 or 64 bytes, but got ${privateKeyBytes.length}`)
   }
   return async (data: string | Uint8Array): Promise<string> => {
     const dataBytes: Uint8Array = typeof data === 'string' ? stringToBytes(data) : data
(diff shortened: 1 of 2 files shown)


src/signers/EdDSASigner.ts
```diff
@@ -18,8 +18,8 @@ import { bytesToBase64url, stringToBytes } from '../util.js'
  */
 export function EdDSASigner(secretKey: Uint8Array): Signer {
   const privateKeyBytes: Uint8Array = secretKey
-  if (privateKeyBytes.length !== 64) {
-    throw new Error(`bad_key: Invalid private key format. Expecting 64 bytes, but got ${privateKeyBytes.length}`)
+  if (![32, 64].includes(privateKeyBytes.length)) {
```
nickreynolds, line 21: any chance you could add a comment that points to a resource explaining that both key lengths are acceptable?

nickreynolds (approved):

src/signers/EdDSASigner.ts
mirceanis: Of course

PR #302, opened 2023-10-31
fix(deps): downgrade uint8arrays and multiformats

### LINKED ISSUE #301: [BUG][deps] the update of uint8arrays>3.x and multiformats>9.x caused broken builds in CommonJS projects
As explained by @nklomp  in https://github.com/decentralized-identity/did-jwt-vc/issues/135, a dependency update was made that cascaded into broken builds.

This library is still being built as both CJS and ESM, but these 2 dependencies stopped publishing CJS variants in recent versions.
Unfortunately the maintainers of the above libraries don't want to publish CJS variants of their libraries so one of the few options remaining is to downgrade those dependencies here to ameliorate the annoyance for users of this library.

The JS world seems to be moving toward ESM but it is painfully slow and annoyingly difficult to support both.
I'm posting this issue here to gauge the support for the downgrade.

👍 downgrade   /  👎 do nothing ?

mirceanis:
fixes #301

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,8 +94,9 @@
     "@scure/base": "^1.1.3",
     "canonicalize": "^2.0.0",
     "did-resolver": "^4.1.0",
-    "multiformats": "^12.0.0",
-    "uint8arrays": "^4.0.3"
+    "multibase": "^4.0.6",
+    "multiformats": "^9.6.2",
+    "uint8arrays": "3.1.1"
   },
   "eslintIgnore": [
     "*.test.ts"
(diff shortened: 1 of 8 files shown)


robdefeo:
I'm blocked by this, and going to have to result in a really ugly patch unless this PR can get merged. Is there a timescale on this being merged and published?

mirceanis:
Support for this fix seems to be gathering fast so perhaps tomorrow unless there's opposition

robdefeo:
Thanks for turning this around so quickly

PR #305, opened 2024-01-17
feat: add support for Multikey

### LINKED ISSUE #304: [proposal] add support for Multikey
Some DIDDocuments can use `Multikey` as a verification method type. The key type must then be inferred from the `multibase(multicodec(keyBytes))` encoding found in the `publicKeyMultibase` property of the verification method.

**Describe the solution you'd like**
It should be possible to verify tokens whose issuer DID document contain `Multikey` verification methods at least pointing to the key types that are already supported.

**Describe alternatives you've considered**
Until #234 is fixed, the verification logic is hardcoded, so a fix must be provided here.

**Additional context**
https://www.w3.org/TR/vc-data-integrity/#multikey
https://identity.foundation/peer-did-method-spec/#resolving-a-didpeer2
https://w3c-ccg.github.io/did-method-key/#document-creation-algorithm

mirceanis:
fixes #304

This PR adds support for `Multikey` verification methods, both for signature verification and for key agreement.
I also added a few internal mappings for known key types, verification methods, algorithms and multiformat codecs.

BREAKING CHANGE: The return types have changed for  of `extractPublicKeyBytes()` and `multibaseToBytes()` from Uint8Arrays to Objects containing the `keyBytes: Uint8Array` and a decoded or inferred `keyType: string | undefined`

Merging this PR should trigger a major version bump

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -3,6 +3,9 @@ import { x25519 } from '@noble/curves/ed25519'
 import type { EphemeralKeyPair } from './encryption/types.js'
 import { varint } from 'multiformats'
 import { BaseName, decode, encode } from 'multibase'
+import type { VerificationMethod } from 'did-resolver'
+import { secp256k1 } from '@noble/curves/secp256k1'
+import { p256 } from '@noble/curves/p256'
 
 const u8a = { toString, fromString, concat }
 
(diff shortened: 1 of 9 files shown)


src/util.ts
```diff
@@ -102,23 +280,25 @@ export function bytesToMultibase(
...
   }
 
   // then assume multicodec, otherwise return the bytes
   try {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     const [codec, length] = varint.decode(bytes)
-    return bytes.slice(length)
+    const possibleCodec: string | undefined =
+      Object.entries(supportedCodecs).filter(([, code]) => code === codec)?.[0][0] ?? ''
+    return { keyBytes: bytes.slice(length), keyType: CODEC_TO_KEY_TYPE[possibleCodec as KNOWN_CODECS] }
```
mirceanis, line 298: This is the most relevant change

nickreynolds (approved):

Issue #234, opened 2022-06-17
Add support for providing verifier


sondreb:
It is currently possible to specify a custom signer, but it's not possible to specify a custom verifier.

`createJWT` allows signing with custom algorithm, like "SS256K" (Schnorr), but it's not possible to specify the verifier using the `verifyJWT`, so if the JWT/JWS is signed with a custom signer, the verifier will fail.

sondreb:
We have in our identity wallet, previously migrated to [@noble/secp256k1](https://github.com/paulmillr/noble-secp256k1), so to avoid dependency on yet another library for the same purpose, I wanted to create our own Signer and Verifier using that library. It's not super important to use the new library, though it's very clean, zero dependencies and very fast.

did-jwt relies on `elliptic` package (doesn't appear to be any development on it for year and a half) which does not implement BIP0340 (Schnorr signatures), so adding support for "SS256K" wouldn't be possible without doing something with dependencies.

Couple of unrelated things I noticed:

- The example in README.md uses `publicKey` instead of `verificationMethod`.
- There is no support for `JsonWebKey2020` (and `JsonWebSignature2020`), but need to use `EcdsaSecp256k1VerificationKey2019`.

My preferred option would be to use JsonWebKey2020 with an Schnorr Public Key, which only defines the x coordinate. Guess that could be marked by setting the "crv" on the jwk field to "SS256K", mapping that to the exact same value as the "alg" on the signed JWT/JWS. I already had a working implementation using "publicKeyMultibase", but was not sure if JWK was a "better" option.

sondreb:
https://github.com/decentralized-identity/did-jwt/commit/60987e0025b1fd45e4a8e583a8c1e28df403fa0b fixes what I mentioned here with no support for `JsonWebKey2020`. Thanks!

PR #317, opened 2024-11-12
feat: Add support for custom signing algorithms

AntonioAlarcon32:
Some use cases need signing algorithms that aren't available in the library. This change makes the library more flexible by letting users add their own algorithms for signing and verification.

This PR adds support for custom signing algorithms beyond the ones currently implemented in the library. The changes are split into three commits:
1. Library modifications to support custom algorithms.
2. New tests for coverage of new code.
3. Documentation updates for implementing custom algorithms.

The changes modify the existing algorithms Record to support customization, adding some functions to add the signing algorithm. No new dependencies are required

If needed, we have created a working example using EIP-712 (Typed structured data hashing and signing) here: 
https://github.com/AntonioAlarcon32/did-jwt-eth-typed-data-signature. Using EIP-712 allows signing of JWTs using wallets like Metamask without any DID document modification, which more coveniently couples DID management with Ethereum accounts.

### DIFF
--- a/docs/guides/index.md
+++ b/docs/guides/index.md
@@ -177,3 +177,55 @@ async function mySigner(data: Uint8Array | string): Promise<string> {
 Your function must returns a `Promise<string>`.
 
 A successful call resolves to a `base64url`-encoded signature.
+
+### Adding signing algorithms
+
+New signing algorithms can be created by implementing the SignerAlgorithm and Signer functions for the signing process and the verify function with the valid signatures.
+
+#### Signing process
+
+Implement a Signer Algorithm function, that returns a function that receives payload and signer and returns the signature as `Promise<string>`
+
+
+```typescript
+async function customSigner(data: Uint8Array | string): Promise<string> {
+  const signatureBytes = await customSignerLogic(data)
+  return bytesToBase64url(signature)
+}
+```
+
+```typescript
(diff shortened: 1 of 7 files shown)


mirceanis:
Thank you for the contribution!

There are some open discussions about customizing the verifiers. See #234 #170

Adding signers is no longer necessary as you can specify the signer when creating the JWT.

For adding verifiers, we shouldn't modify the internal mapping globally. Either provide custom verifiers as parameters to verifyJWS / JWT, or adopt a builder pattern to construct a verifier that supports extra algorithms.

juanelas:
Thank you for the prompt reply. I'm working with @AntonioAlarcon32 in the same project.

You point us to the actual support of external signers. However, they are not an option since they can be only used for the algorithms that are already hardcoded internally, and we need to use different ones, such as EIP-712.

On your [Jul 5 2022 comment](https://github.com/decentralized-identity/did-jwt/issues/170#issuecomment-1174939528) to issue #170 you considered removing the internally hardcoded mapping of algorithms to signers and verifiers. This would be in our option a good approach. One way to address it would be to create a standard interface/class JWTSigner supporting one or more algorithms that are able to sign with, and something similar with verifiers. In order to keep backwards compatibility (no breaking changes), we think we could support both the actual Signer signature and the new JWTSigner.

Would you consider such an approach for merging? If it is case, we could start developing it promptly.

Thank you in advance for your time!

PR #337, opened 2026-08-18
fix: reject non-NumericDate exp, nbf, and iat claims

SashaMIT:
## Summary
- `verifyJWT` treated `exp`, `nbf`, and `iat` as present only when they were truthy. `exp: 0` and string values such as `"never"` skipped the RFC 7519 NumericDate window while the signature still verified.
- Fail closed when those claims are present but not a finite number. `exp: 0` is expired (1970-01-01). Missing claims stay optional, matching the existing policy flags.
- Pin `policies.now` on the hardcoded did:key VC. That fixture expired on 2025-12-04 and already failed on tip.

## Test plan
- [x] `yarn test` 326/326
- [x] Revert-test: the four new assertions fail when the guards are removed
- [x] `yarn lint`

Threat-model: the signer already controls the payload. The remaining verifier guard is the validity window. A present non-NumericDate claim turned that guard into a no-op.

### DIFF
--- a/src/JWT.ts
+++ b/src/JWT.ts
@@ -161,6 +161,14 @@ function encodeSection(data: any, shouldCanonicalize = false): string {
 
 export const NBF_SKEW = 300
 
+function isNumericDate(value: unknown): value is number {
+  return typeof value === 'number' && Number.isFinite(value)
+}
+
+function hasTimeClaim(value: unknown): boolean {
+  return value !== undefined && value !== null
+}
+
 function decodeJWS(jws: string): JWSDecoded {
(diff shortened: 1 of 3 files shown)


mirceanis:
@SashaMIT Thank you for the contribution!
I broke CI on this PR by attempting to fix merge conflicts through the GH interface (test snapshots became outdated).

I'm pushing #344 with the fixed merge and a changeset included, with your original commit included.

SashaMIT:
Thanks for landing this, and for pulling the original commit through after the merge-conflict snapshot break. Glad exp, nbf, and iat now fail closed on a non-NumericDate.

PR #348, opened 2026-09-19
test: remove mockdate dependency

whydrae:
This PR removes the `mockdate` dependency and replaces it with the vitest native date mocks.


mirceanis (approved):
Looks good, thanks!

PR #349, opened 2026-09-19
test: remove tweetnacl dependency

whydrae:
This PR removes the `tweetnacl` (which is not maintained for a long time) and replaces it with the  `@noble/curves` functionality.


mirceanis (approved):
Looks good.

Tweetnacl was kept around as an oracle to check interoperability with `@noble` libs. A pure js cryptography library doesn't have many reasons to be maintained, so I didn't see a problem with it. It's not in the runtime dependencies anyway.

I'll merge this nonetheless. Thank you for the contribution!

PR #350, opened 2026-09-19
test: remove jwk-to-pem dependency

whydrae:
Another small dev dependency cleanup: the `jwk-to-pem` can be also safely replaced with the native Node functions. Even thought it's dev dependency, it also has some unmaintained dependencies (`elliptic`, `asn1.js`).


mirceanis (approved):
looks good. thanks!

